PR #18, opened 2022-10-24
Fix edge cases relating to running Delete in a partial EGC.

xorgy:


xorgy:
This is better than the existing solution, but only marginally so.

jackpot51 (approved):

PR #19, opened 2022-10-25
Fixed some Clippy issues, Rustfmt and orbclient scroll/select fix

genusistimelord:
Fixed some Clippy issues, Rustfmt and made it so the editor-orbclient scrolls and selects like a real textbox.

### DIFF
--- a/examples/editor-orbclient/src/main.rs
+++ b/examples/editor-orbclient/src/main.rs
@@ -26,7 +26,7 @@ fn main() {
         1024 * display_scale as u32,
         768 * display_scale as u32,
         &format!("COSMIC TEXT - {}", font_system.locale),
-        &[WindowFlag::Resizable],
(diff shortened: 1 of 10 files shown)


genusistimelord:
closing due to rustfmt.

examples/editor-orbclient/src/main.rs
```diff
@@ -26,7 +26,7 @@ fn main() {
         1024 * display_scale as u32,
         768 * display_scale as u32,
         &format!("COSMIC TEXT - {}", font_system.locale),
-        &[WindowFlag::Resizable],
+        &[WindowFlag::Resizable, WindowFlag::Async],
```
jackpot51, line 29: Why are you enabling async? That will spin even when there are no events.

jackpot51:
I'm going to revert some of the formatting changes and squash that into your commit.

jackpot51:
Replaced by #20

PR #20, opened 2022-10-25
Clippy fixes and editor-orbclient scroll drag selection fixes and middle mouse scroll fixes

genusistimelord:
the scroll bar is no longer tied to the cursor. It now also only re-renders on event or when the mouse is held down and withing 5 pixels of the top or bottom during selection.

### DIFF
--- a/examples/editor-orbclient/src/main.rs
+++ b/examples/editor-orbclient/src/main.rs
@@ -98,19 +98,15 @@ fn main() {
     let mut mouse_x = -1;
     let mut mouse_y = -1;
     let mut mouse_left = false;
-    let mut shape_me = true;
+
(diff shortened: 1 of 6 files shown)


examples/editor-orbclient/src/main.rs
```diff
@@ -98,11 +98,19 @@ fn main() {
...
+
     loop {
         let font_size = buffer.metrics().font_size;
         let line_height = buffer.metrics().line_height;
 
-        buffer.shape_until_cursor();
+        let mut force_drag = true;
+
+        //lets do this once and rely on the events to do the rest.
+        if shape_me {
```
jackpot51, line 110: This can move outside of the loop if it is only done once

jackpot51 (approved):

PR #25, opened 2022-10-28
Rework add_span

genusistimelord:
Rework Ranges to split and change or overwrite based on old ranges to new range within add span

a playground test of it.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dc84119d7c4ca64743a27b947b14094f


13r0ck:
If you dont use `-m` on `git commit` you can break you commit messages into multiple lines, so they can still be verbose (thank you for that!), but also formatted correctly

genusistimelord:
> If you dont use `-m` on `git commit` you can break you commit messages into multiple lines, so they can still be verbose (thank you for that!), but also formatted correctly

thank you!

jackpot51 (approved):

PR #32, opened 2022-11-03
Add links to repo and docs in Cargo.toml

WhyNotHugo:
This will add a link to the repository in docs.rs, and both links to crates.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,8 @@ version = "0.3.0"
 authors = ["Jeremy Soller <jeremy@system76.com>"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -5,6 +5,8 @@ version = "0.3.0"
 authors = ["Jeremy Soller <jeremy@system76.com>"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
+documentation = "https://pop-os.github.io/cosmic-text/cosmic_text/"
```
jackpot51, line 8: This doc link is for the master branch, would it be better to link to docs.rs here so the crate references only the documentation of released versions?

WhyNotHugo: You're right, fixed.

jackpot51 (approved):

PR #34, opened 2022-11-04
bugfix: text selection starting from the bottom

hojjatabdollahi:
This fixes a bug in the text selection.
Steps to reproduce the bug:
- Have a paragraph of text.
- Select a portion of that text.
- Then try dragging from below the last line (where there is no text)
- The selected text instead of starting from the bottom, starts from the end of previous selection to the current mouse position.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -477,6 +477,16 @@ impl<'a> Buffer<'a> {
                 break;
             }
         }
+        // Fixes text selection if the user starts dragging from below the last line.
+        if let Some(run) = self.layout_runs().last() {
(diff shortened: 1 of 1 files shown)


jackpot51 (changes requested):
For more efficiency, not having to go through all layout runs again, consider storing what you need from the last run in the previous layout_runs loop

hojjatabdollahi:
@jackpot51 Would switching to a peekable iterator be satisfactory?

I need to know what is the y for the last line.

jackpot51 (approved):
Looks good to me, thanks!

PR #36, opened 2022-11-07
Add Padding to the TextBox

hojjatabdollahi:
This adds text margins to the TextBox and makes sure text selection works as expected.
I'm leaving this PR here in case you guys have any suggestions.

- [x] `left_margin` and `right_margin`.
- [x] Fix the issue with text selection if you start dragging from the side margins.
- [x] `top_margin` and `bottom_margin`.
- [x] Text selection has an issue (in LTR only) when selecting from the beginning of the line until the left half of the first glyph. It erroneously selects the whole line.
- [x] Text selection has an issue when selecting text in a RTL paragraph that contains LTR segments. Specifically when the LTR segment falls on a separate line. The last letter in the LTR segment can't be selected.


jackpot51:
This needs to be rebased on recent changes

hojjatabdollahi:
@jackpot51  I added the padding to the SyntaxTextBox too. But the theme colors the `editor` differently than the `SyntaxTextBox` container (outside the editor rectangle, but inside the TextBox).

But I noticed that SyntaxTextBox has a few issues with themes (for example light theme doesn't work) or switching monospaced off/on would also break the theme.

jackpot51:
#40 will merge SyntaxTextBox and TextBox. It is intentional for SyntaxTextBox to ignore colors from the stylesheet, since the syntax highlighter has its own theming.

jackpot51:
Ideally the editor-libcosmic code would switch the SyntaxEditor theme to a light theme as needed.

PR #39, opened 2022-11-08
Remove `Mutex` from `FontSystem`, and make methods take `&mut self`

ids1024:
Types wrapping `&FontSystem` instead need to have methods taking an `&mut FontSystem`. Eliminating the lifetime bounds make the API more usable, but perhaps too many methods require this to be passed.

Moving interior mutability to the caller means it isn't used unless needed, and the caller can use a `Mutex`, some nostd alternative, or a `RefCell`.

This updates the `editor-orbclient` to compile with the API changes. The others need to be updated, but perhaps not until any changes to make methods not require an `&mut FontSystem` if possible.


hecrj (approved):
Getting rid of the `Buffer` lifetime is great. It'll help me get rid of `ouroboros` in [my `iced` integration](https://github.com/iced-rs/iced/blob/16b215138ea3a48703f7e83be6d73ed6ca9af465/wgpu/src/text.rs#L22-L33).

tigregalis:
> Types wrapping `&FontSystem` instead need to have methods taking an `&mut FontSystem`. Eliminating the lifetime bounds make the API more usable, but perhaps too many methods require this to be passed.

Perhaps as future work, to recover some of the ergonomics, when you are calling methods on the same instance repeatedly, so that you're not always passing `&mut font_system` to every method, you could expose a helper type that borrows the inner type and the `FontSystem` then forwards the methods:

```rust
/// Gets dropped at the end of the scope, releasing the borrows
struct WithFontSystem<'a, T> {
    inner: &'a mut T,
    font_system: &'a mut FontSystem,
}

/// Use like
/// ```
/// let mut buffer = Buffer::new();
/// let param = ();
/// let mut buffer = buffer.with_font_system(&mut font_system);
/// buffer.method(param);
/// ```
impl Buffer {
    fn method(&mut self, font_system: &mut FontSystem, param: Param) {
        todo!()
    }

    fn with_font_system<'a>(&'a mut self, font_system: &'a mut FontSystem) -> WithFontSystem<'a, Self> {
        WithFontSystem {
            inner: self,
            font_system,
        }
    }
}

impl<'a> WithFontSystem<'a, Buffer> {
    // ... reimplement all of the methods on `Buffer`, simply forwarding to the relevant `Buffer` methods
    fn method(&mut self, param: Param) {
        self.inner.method(&mut self.font_system, param);
    }
}
```

geieredgar:
@ids1024 @tigregalis Thank you for your ideas, which I have incorporated in #97. With #97 being merged, I think this can be closed?

PR #40, opened 2022-11-10
Vi-style editor and other editor improvements

jackpot51:
When merged, will release version 0.5


genusistimelord:
this looks nice. I did however push a PR for changing the attr list to be a Btree now. Please review and maybe we can merge it into 0.5? I see it overlaps some changes here.

PR #43, opened 2022-11-18
🍮 Add custom font support

Jasper-Bekkers:
This addresses (at least in the `no_std` case, I can amend the `std` case as well) the todo's that were left about being able to load custom fonts. We needed this for our use-case since we can't rely on OS fonts in our case.

### DIFF
--- a/src/font/mod.rs
+++ b/src/font/mod.rs
@@ -29,8 +29,6 @@ impl<'a> Font<'a> {
             }
             #[cfg(feature = "std")]
             fontdb::Source::SharedFile(_path, data) => data.deref().as_ref(),
-            #[cfg(not(feature = "std"))]
-            _ => { todo!(""); }
         };
 
         Some(Self {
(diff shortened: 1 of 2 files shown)


src/font/mod.rs
```diff
@@ -29,6 +29,8 @@ impl<'a> Font<'a> {
             }
             #[cfg(feature = "std")]
             fontdb::Source::SharedFile(_path, data) => data.deref().as_ref(),
+            #[cfg(not(feature = "std"))]
+            _ => { todo!(""); }
```
Jasper-Bekkers, line 33: This was needed as a compile-fix for the no_std build.

jackpot51: What other enum variant is there?

Jasper-Bekkers: None but 2 of the variants aren't handled in the std case, so I'm defaulting to _ here mostly out of convenience.

Jasper-Bekkers: Looks like my bad, I accidentally used `fontdb` normally as well as one of my dependencies!

jackpot51 (approved):

PR #45, opened 2022-11-21
❎ Add cargo-deny to the CI of this crate

Jasper-Bekkers:
This adds [cargo deny](https://github.com/EmbarkStudios/cargo-deny-action) support to the CI steps of this repository. This should make it easier to prevent duplicate dependencies and to detect security warnings or license violations going forward.

I added this because ttf-parser got brought in twice by this crate; once because of the dependency on `fontdb` and once because of the dependency on `rustybuzz`.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,269 @@
+# This template contains all of the possible sections and their default values
+
+# Note that all fields that take a lint level have these possible values:
+# * deny - An error will be produced and the check will fail
+# * warn - A warning will be produced, but the check will not fail
+# * allow - No warning or error will be produced, though in some cases a note
+# will be
+
+# The values provided in this template are the default values that will be used
+# when any section or field is not specified in your own configuration
+
+# Root options
+
+# If 1 or more target triples (and optionally, target_features) are specified,
+# only the specified targets will be checked when running `cargo deny check`.
+# This means, if a particular package is only ever used as a target specific
+# dependency, such as, for example, the `nix` crate only being used via the
+# `target_family = "unix"` configuration, that only having windows targets in
+# this list would mean the nix crate, as well as any of its exclusive
+# dependencies not shared by any other crates, would be ignored, as the target
+# list here is effectively saying which targets you are building for.
+targets = [
+    # The triple can be any string, but only the target triples built in to
+    # rustc (as of 1.40) can be checked against actual config expressions
+    #{ triple = "x86_64-unknown-linux-musl" },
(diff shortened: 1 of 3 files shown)


deny.toml
```diff
@@ -0,0 +1,269 @@
...
+# A list of advisory IDs to ignore. Note that ignored advisories will still
+# output a note when they are encountered.
+ignore = [
+    # = Advisory: https://rustsec.org/advisories/RUSTSEC-2022-0048
+    # = xml-rs is a XML parser has open issues around parsing including integer
+    #     overflows / panics that may or may not be an issue with untrusted data.
+
+    #     Together with these open issues with Unmaintained status xml-rs
+    #     may or may not be suited to parse untrusted data.
+    "RUSTSEC-2022-0048",
```
Jasper-Bekkers, line 83: Currently this crate brings in `xml-rs` which has a security advisory against it, though I switching xml library was beyond the scope of this PR so I chose to just ignore the advisory for now and highlight it in this PR instead

deny.toml
```diff
@@ -0,0 +1,269 @@
...
+# The lint level for crates which do not have a detectable license
+unlicensed = "deny"
+# List of explicitly allowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+allow = [
+    "MIT",
+    "Apache-2.0",
+    "Unicode-DFS-2016",
+    "BSD-2-Clause",
```
Jasper-Bekkers, line 114: These are the licenses currently used by this library, so I chose to include all of them here.

jackpot51 (approved):

MarijnS95:
https://github.com/pop-os/cosmic-text/actions/runs/5229814267/jobs/9443023522 it looks like these duplicates are only treated as warnings and the same `ttf-parser` clash is there again: https://github.com/RazrFalcon/rustybuzz/pull/69

MarijnS95:
Looks like that was "intentional": https://github.com/pop-os/cosmic-text/commit/587bb5c98ff4ce0a5aa123b72c40d6fa3fffb335

jackpot51:
I would want to have this resolved by the next cosmic-text release. I'll keep an eye on when rustybuzz gets a new release.

jackpot51:
Related: https://github.com/trishume/syntect/pull/476

MarijnS95:
@jackpot51 looks like both dependencies had their bump merged now, and `rustybuzz` already made it into a release. Now the wait is for `syntect`.

PR #48, opened 2022-11-30
Use the `fontdb` matching algorithm instead of a custom one

Jasper-Bekkers:
This code is quite WIP and shouldn't be merged in as-is.

Having said that, I would like to know what people's thoughts are around this. Currently it appears that `cosmic-text` is trying to find an exact match to the font that's currently required.

On the other hand, fontdb implements the CSS font-matching algorithm which seems to lead to better results for us when doing more fuzzy matching. (For example; searching for a bold font but not finding one of the exact weight will find one with the closes weight, etc).

If that's a way to go, I'd love to polish up this PR.

### DIFF
--- a/src/font/system/no_std.rs
+++ b/src/font/system/no_std.rs
@@ -61,15 +61,24 @@ impl FontSystem {
 
     pub fn get_font_matches<'a>(&'a self, attrs: Attrs) -> Arc<FontMatches<'a>> {
         let mut fonts = Vec::new();
-        for face in self.db.faces() {
-            if !attrs.matches(face) {
-                continue;
-            }
+        // for face in self.db.faces() {
+        //     log::info!("{:?}", face.family);
+        //     if !attrs.matches(face) {
+        //         continue;
+        //     }
 
-            if let Some(font) = self.get_font(face.id) {
-                fonts.push(font);
+        //     if let Some(font) = self.get_font(face.id) {
+        //         fonts.push(font);
+        //     }
+        // }
+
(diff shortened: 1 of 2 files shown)


src/font/system/no_std.rs
```diff
@@ -61,15 +61,24 @@ impl FontSystem {
 
     pub fn get_font_matches<'a>(&'a self, attrs: Attrs) -> Arc<FontMatches<'a>> {
```
jackpot51, line 62: This function has to get _all_ fonts that match attrs for font fallback to work

Jasper-Bekkers: I'm not familiar enough with this system to make that work; but it looks like there is some interest in this PR from various folks contributing to this project. Maybe somebody can pick up where I've left off?

jackpot51: If that is the case, I will close the PR and hopefully a compliant implementation will be done at some point.

Jasper-Bekkers: I just realized; `self.db.query` already takes care of the font fallback here and thus this only needs to return one font going forward. We just need to build up the query so that it's always containing something that matches.

jackpot51: `db.query` absolutely does not do font fallback. To do so requires shaping with the default font, then finding missing glyphs, then shaping those with the fallback font list.

jackpot51: (Which is what the code in src/font/fallback in cosmic-text does)

PR #50, opened 2022-12-15
Add line width field to `LayoutLine`

hecrj:
Added a `w` field to `LayoutLine` and a `line_w` to `LayoutRun`.

These are useful to avoid iterating through all the glyphs just to figure out the line width (e.g. for alignment purposes).

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -55,6 +55,8 @@ pub struct LayoutRun<'a> {
     pub glyphs: &'a [LayoutGlyph],
     /// Y offset of line
     pub line_y: i32,
+    /// width of line
+    pub line_w: f32,
 }
 
 /// An iterator of visible text lines, see [`LayoutRun`]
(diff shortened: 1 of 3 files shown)


jackpot51:
Does this work with RTL text?

hojjatabdollahi:
@jackpot51  Just FYI, I'm redoing the RTL logic. As currently it's not wrapping correctly. It would change the `ShapeLine::layout ` function considerably.

jackpot51:
> This won't work for RTL.
> 
> @jackpot51 Just FYI, I'm redoing the RTL logic. As currently it's not wrapping correctly. It would change the `ShapeLine::layout ` function considerably.

Sounds good to me

hecrj:
> This won't work for RTL.

Tried implementing a fix in 8b665c310334a72e22806696ef1e1521690b1d24.

jackpot51 (approved):

PR #52, opened 2022-12-15
Expose subpixel offsets in `LayoutGlyph`

hecrj:
This PR exposes the `x_offset` and `y_offset` of a `LayoutGlyph` instead of just the integer components.

I believe these are necessary to properly position glyphs when dealing with logical units instead of physical ones (e.g. when rendering outlines #51).

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -20,6 +20,10 @@ pub struct LayoutGlyph {
     pub rtl: bool,
     /// Cache key, see [CacheKey]
     pub cache_key: CacheKey,
+    // X offset in line
+    pub x_offset: f32,
+    // Y offset in line
+    pub y_offset: f32,
     /// Integer component of X offset in line
     pub x_int: i32,
     /// Integer component of Y offset in line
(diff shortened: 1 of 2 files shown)


jackpot51:
Fractional offsets can be obtained from cache_key x_bin and y_bin

hecrj:
Wouldn't that produce an error since the binnings are discrete? Couldn't the errors become noticeable when scaling the layout (e.g. DPI, `Canvas` scale transformation, etc.).

src/layout.rs
```diff
@@ -20,6 +20,10 @@ pub struct LayoutGlyph {
     pub rtl: bool,
     /// Cache key, see [CacheKey]
     pub cache_key: CacheKey,
+    // X offset in line
```
jackpot51, line 23: If the docs could explain the difference between using this and x_int/y_int it would be good. Basically, if you want to cache with the CacheKey, you should use the x_bin/y_bin there.

hecrj: I tried to add some clarification in c46daa150106c070b1691d2bff8ffbd1a02f0e1f.

jackpot51 (approved):

PR #54, opened 2022-12-17
Bidi reordering and wrapping

hojjatabdollahi:
Previously we were calling `unicode_bidi::bidi_info::visual_run` to reorder the embedded bidi text. This was working well for texts on one line, but wrapping text would break the algorithm.

I had to move the reordering algorithm into `cosmic_text` to be able to adapt it. Also I had to change the `ShapeLine::layout` function to create the `LayoutLine` at the end, after reordering.

Currently I can not do `wrap_simple` because I am using a `Vec<span_index, word_range>` to represent a line. But to be able to do `wrap_simple` I need to know where in a word the line break happens.

More detail: 
So with this PR, if we have a big chunk of LTR text that needs multiple visual lines. It would become one single `Span` with index `0` and the visual lines would be `[ [(0, 0..15)], [(0, 15..34)], ..]` (which is a `Vec<Vec<usize, Range>>`). Now if we have two small chunks of text, one 5 words LTR, one 4 words RTL and it would fit on one line. It would generate two Spans (indices `0`,`1`) and the Visual line would be `[[(0, 0..5), (1, 0..4)]]`. 

TODO:
- [x] Correct BiDi ordering per visual line
- [x] wrap_simple
- [x] Skip first space on a new line
- [x] A long sequence of whitespaces is considered a word and causes issues with wrapping
- [x] Words longer than the `line_width` break the layout (in the main branch too).

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -711,6 +711,10 @@ impl ShapeLine {
                             }
                             fitting_start = i;
                             fit_x = start_x;
+                            if i >= span.words.len() {
+                                break;
+                            }
+                            continue;
                         }
 
                         if self.rtl {
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -651,6 +651,9 @@ impl ShapeLine {
 
                         if wrap {
                             let mut fitting_start = i + 1;
+                            if fitting_start == fitting_end { //long single word
+                                fitting_start -= 1;
+                            }
                             while fitting_start < fitting_end {
                                 if span.words[fitting_start].blank {
                                     fitting_start += 1;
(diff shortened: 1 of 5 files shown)


src/shape.rs
```diff
@@ -276,19 +278,20 @@ impl ShapeGlyph {
 pub struct ShapeWord {
     pub blank: bool,
     pub glyphs: Vec<ShapeGlyph>,
+    x_advance: f32,
```
jackpot51, line 281: Make these pub?

hojjatabdollahi: I made them public.

hojjatabdollahi:
Currently (in the main) we shape a span like this: `"Hello␣␣␣␣␣␣Bye"` into 3 `WordSpans` like this: `["Hello␣␣␣␣␣", "␣", "Bye"]`.
This messes up wrapping. Since the first word is considered a 10 letter word (instead of 5 letter word "hello").
In the commit [0ac4763](https://github.com/pop-os/cosmic-text/pull/54/commits/0ac4763f8d1ec76addb87c2c5cca714c36d18e22), I'm turning all the whitespaces into a separate word. So this example would become: `["Hello", "␣", "␣", "␣", "␣", "␣", "␣", "Bye"]`.

This would affect efficiency. But the result is correct (at least as much as I could test).

hojjatabdollahi:
The latest commit would handle very long words. It will render only the visible parts. The caret would leave the screen if you keep typing a long word (longer than the width of the textbox), a horizontal scroll will handle this.

Simple_wrap should break these type of words. I haven't finished simple_wrap yet.

jackpot51:
Good work so far! Let me know when you want a review.

hojjatabdollahi:
@jackpot51 I think it's ready for a review.

jackpot51 (approved):

jackpot51:
This looks like good improvements to bidi layout, I will release it soon.

Issue #57, opened 2022-12-27
Loading fonts incrementally


rhelmot:
I see #43 added support for constructing a FontSystem with custom fonts loaded, but is there a way to get a mutable reference to the font database after the system is constructed so that more fonts can be loaded afterwards?

rhelmot:
From reviewing the source code it seems the way the caching works and the way orobouros is used prohibits this. Is this something that might change at any point in the future? externally managed caches? #39?

PR #60, opened 2022-12-30
Many changes to support using cosmic-text with existing widget and rendering pipeline

rhelmot:
I'm a developer on [Vizia](https://github.com/vizia/vizia) and we're trying to replace our existing text rendering pipeline (using femtovg's basic vector-based text rendering) and homemade text editor with cosmic-text. This is a giant pile of changes I had to make in order to make everything work correctly and reach feature parity. If you like, I can break this into multiple PRs, but it will be difficult.

- 1118631cfb5e4b3b49fbc753ec01413956268fb4 and 5623d11b1614859b702146e0cbb6f32652ce138f are workarounds for #57. Being able to tear down existing structures and rebuild them without cloning data is useful. If people like, I can share the snippet which lets me load fonts incrementally.
- 866e4e33c3c384ed994e033c4442046218511016 and https://github.com/pop-os/cosmic-text/commit/64e33faac8ee8e87d8bb83ac09333503e4a95104 are a port of #43 to the `std` font system
- 23afcaa033fbf06006b930fe7d24fca4e9223787 re-exports fontdb. Pending something like #48, this is necessary to be able to use fontdb's query algorithm to perform queries without having to keep our fontdb dependency synchronized with cosmic.
- 76660679a22607e1b235f22de6cc7243571014bc answers the question "how many wrap-rendered lines are there in a buffer", which is important for performing layout when height is dependent on text.
- 9876679a1699181cee991857e1d8abe29766efc7, 6582ce5ff31f34ce51a50b3504e213206f19ce8d, 52533be0fd9b864b8ae1f1566a6f27b21d1c5c65, and 3e355f0cfc9cebcf958c88c48cbe12a3ac672ac7 add Editor::insert_text, which lets you insert multiple characters or even multiple lines at once.
- 6582ce5ff31f34ce51a50b3504e213206f19ce8d in particular also adds several new editor Action verbs as well as fixing the PageUp and PageDown actions.
- 96a68dfa4aa99ced08733939459ff3287aae2647 adds a method on Buffer to get the per-line bounding boxes of a given cursor range. I'm using this to render highlights and cursors. This could be integrated into the Editor::draw method, but I didn't want to mess up anything that was depending on the draw order it provides.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -64,16 +64,26 @@ pub struct LayoutRunIter<'a, 'b> {
     buffer: &'b Buffer<'a>,
     line_i: usize,
     layout_i: usize,
+    remaining_len: usize,
     line_y: i32,
     total_layout: i32,
 }
 
 impl<'a, 'b> LayoutRunIter<'a, 'b> {
     pub fn new(buffer: &'b Buffer<'a>) -> Self {
+        let total_layout_lines: usize = buffer.lines.iter().map(|line| line.layout_opt().as_ref().map(|layout| layout.len()).unwrap_or_default()).sum();
+        let top_cropped_layout_lines = total_layout_lines.saturating_sub(buffer.scroll.try_into().unwrap_or_default());
+        let maximum_lines = buffer.height.checked_div(buffer.metrics.line_height).unwrap_or_default();
+        let bottom_cropped_layout_lines = if top_cropped_layout_lines > maximum_lines.try_into().unwrap_or_default() {
+            maximum_lines.try_into().unwrap_or_default()
+        } else {
+            top_cropped_layout_lines
+        };
         Self {
             buffer,
             line_i: 0,
             layout_i: 0,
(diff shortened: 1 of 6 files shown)


rhelmot:
While attempting to flesh out support for RTL buffers, I found that the Buffer::highlight_blocks interface was the wrong abstraction, and switched it to LayoutLine::highlight. This seems to be working well, and could definitely be the basis for cursor xpos preservation on vertical movement (probably the next thing I will do)

jackpot51:
I will try to look into this sometime this week, thanks for the pull request. I may cherry pick the smaller commits first, and then rebase.

genusistimelord (commented):
This looks pretty good and a lot of thought into this. personally the changes seem small enough code wise that combining them probably will be OK.  I quickly went over the code and it looks fine to me.

jackpot51 (approved):
This looks good and doesn't appear to have breaking changes

PR #62, opened 2023-01-11
Add cursor affinity

rhelmot:
This is a breaking change on account of adding a new public field to the Cursor struct.

However, it does significantly improve the user experience with respect to moving the cursor between soft-wrapped lines. The new behavior doesn't _perfectly_ match the rest of the applications on my operating system (ubuntu) but it seems to make the home/end keys as well as up/down at the start or end of lines much nicer.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -19,12 +19,24 @@ pub struct Cursor {
     pub line: usize,
     /// First-byte-index of glyph at cursor (will insert behind this glyph)
     pub index: usize,
+    /// Whether to associate the cursor with the run before it (false) or the run after it (true)
+    /// if placed at the boundary between two runs
+    pub affinity: bool,
 }
 
 impl Cursor {
     /// Create a new cursor
     pub const fn new(line: usize, index: usize) -> Self {
-        Self { line, index }
+        Self::new_affinity(line, index, false)
+    }
+
+    /// Create a new cursor, specifying the affinity
+    pub const fn new_affinity(line: usize, index: usize, affinity: bool) -> Self {
+        Self {
+            line,
+            index,
+            affinity,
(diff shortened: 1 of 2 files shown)


src/buffer.rs
```diff
@@ -19,12 +19,24 @@ pub struct Cursor {
     pub line: usize,
     /// First-byte-index of glyph at cursor (will insert behind this glyph)
     pub index: usize,
+    /// Whether to associate the cursor with the run before it (false) or the run after it (true)
+    /// if placed at the boundary between two runs
+    pub affinity: bool,
```
nicoburns, line 24: I feel like it would be clearer if this was an enum with `Before` and `After` variants instead of a boolean.

src/buffer.rs
```diff
@@ -19,12 +19,24 @@ pub struct Cursor {
...
 
 impl Cursor {
     /// Create a new cursor
     pub const fn new(line: usize, index: usize) -> Self {
-        Self { line, index }
+        Self::new_affinity(line, index, false)
+    }
+
+    /// Create a new cursor, specifying the affinity
+    pub const fn new_affinity(line: usize, index: usize, affinity: bool) -> Self {
```
nicoburns, line 34: Nit: I think `new_with_affinity` or `with_affinity` would be a more idiomatic name for this method.

nicoburns (changes requested):

rhelmot:
Fixed both.

Side note: I initially reached for #[Derive(Default)] on the enum, but then I realized that you don't actually have a minimum rust version specified in your cargo.toml. Might be nice to add.

rhelmot:
(from testing, it seems the MSRV should be 1.63)

PR #66, opened 2023-01-27
Avoid negative width/height in `Buffer::set_size`

grovesNL:
Negative width/height may cause unexpected effects on the layout, which we can avoid by clamping.


jackpot51 (approved):

grovesNL:
That would probably make sense. I wanted to keep the upper bound at `i32::MAX` instead of `u32::MAX` for now because we use these values in other places as `i32`. We could probably switch to `u32` if we either clamp the `u32` to `i32::MAX` or audit all the places these are used to make sure they could work with `u32` instead.

PR #72, opened 2023-02-10
Use `f32` instead of `i32` for lengths

hecrj:
This allows users to use logical coordinates instead of physical ones.

These changes seem to fit naturally since fewer conversions are necessary during layouting. I left the `editor` module largely unchanged because I am not familiar with it, but I imagine it can be ported to logical units as well.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -192,7 +192,11 @@ impl<'a, 'b> LayoutRunIter<'a, 'b> {
             .sum();
         let top_cropped_layout_lines =
             total_layout_lines.saturating_sub(buffer.scroll.try_into().unwrap_or_default());
-        let maximum_lines = (buffer.height / buffer.metrics.line_height) as i32;
+        let maximum_lines = if buffer.metrics.line_height == 0.0 {
+            0
+        } else {
+            (buffer.height / buffer.metrics.line_height) as i32
+        };
         let bottom_cropped_layout_lines =
             if top_cropped_layout_lines > maximum_lines.try_into().unwrap_or_default() {
                 maximum_lines.try_into().unwrap_or_default()
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -107,7 +107,7 @@ pub struct LayoutRun<'a> {
     /// The array of layout glyphs to draw
     pub glyphs: &'a [LayoutGlyph],
     /// Y offset of line
-    pub line_y: i32,
+    pub line_y: f32,
     /// width of line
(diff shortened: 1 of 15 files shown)


dhardy:
Note that use of an integer type for input has a couple of advantages:

- It implies units are physical pixels. Using `f32` one should specify units (and possibly convert for scale factor somewhere?)
- It allows pixel-level fitting. It is important that output is aligned to the pixel grid.
- It ensure some level of sanity checking and binning of input. (Though not necessarily the right kind of binning: e.g. the difference between 100 and 101 may not be important.)

As an alternative, we could have some fun with [IEEE 754](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) for binning:
```rust
// Reject NaN, inf, negative values; apply *some* maximum
assert!(0.0 <= size && size <= 1e6);
// Bin, rounding to nearest (overflow of the mantissa increments the exponent while keeping an implied 1)
let size = size.to_bits();
let size = (size + 0x0002_0000) & 0xFFFC_0000;
let size = f32::from_bits();
```

hecrj:
(1) is self-evident. The whole point of the changes here is to use logical units.

When it comes to (2) and (3), given we are already aiming for layout linearity (i.e. no metrics hinting!), pixel grid alignment (and binning) should happen during rendering (after scaling!), and not during layout.

jackpot51:
@hecrj there are some conflicts after #78, will you be able to fix them or should I?

hecrj:
@jackpot51 I will try to rebase in a bit!

hecrj:
Alright. Rebased and conflicts fixed!

src/buffer.rs
```diff
@@ -192,16 +192,14 @@ impl<'a, 'b> LayoutRunIter<'a, 'b> {
             .sum();
         let top_cropped_layout_lines =
             total_layout_lines.saturating_sub(buffer.scroll.try_into().unwrap_or_default());
-        let maximum_lines = buffer
-            .height
-            .checked_div(buffer.metrics.line_height)
-            .unwrap_or_default();
+        let maximum_lines = (buffer.height / buffer.metrics.line_height) as i32;
```
hecrj, line 195: We may want to still check here?

jackpot51: Yep, line height could by accident be 0

hecrj: It should be handled in 22b8a534facb41a7a92156def84c7ad7e2614b84.

jackpot51 (approved):

PR #84, opened 2023-02-25
WIP: Ellipsize 

hojjatabdollahi:
This is just the beginning for Ellipsize, and I have questions! I would appreciate any feedback.

Here is the Ellipsize enum, I think multi-line ellipsization would only be useful for `Ellipsize::End` :
```rust

pub enum Ellipsize {
    /// No Ellipsize
    None,
    /// Ellipsis at the start, will create a single line with ellipsis at the start
    Start,
    /// Ellipsis in the middle, will create a single line with ellipsis in the middle
    Middle,
    /// Ellipsis at the end, can create one, or more lines depending on the `HeighLimit` and `Wrap`
    End(HeightLimit),
}

pub enum HeightLimit {
    /// Default is a single line
    Default,
    /// Number of maximum lines allowed, `Line(0)`, or `Line(1)`, would be the same as `Default`,
    /// which is a single line. The ellipsis will show at the end of the last line.
    Lines(u8),
    /// The amount of text is limited to the current buffer height. Depending on the font size,
    /// fewer or more lines will be displayed.
    Height,
}
```

Current implementation:
- I'm shaping a "…" character to match the font family and size and store it in the ShapeLine to insert it later where it's needed. This will be customizable.
- Currently I store the ellipsize attribute in the `Text`. I think it should be stored in any widget that uses `BufferLine`. For example a label can have multiple paragraphs, but if I want it to ellipsize after 5 lines, it should happen regardless of how many paragraphs fit in those 5 lines. I doubt that we want to have ellipsizing per paragraph. 

Questions:
1- Do we want ellipsis based on height? Depending on the font size you might be able to fit more or fewer lines in the same text area.

2- When it comes to text selection, what should happen? If I create a label with some of the text ellipsized, and then select the text (including the ellipsis) and copy it, does that mean I copy the ellipsis character? or the hidden text?

### DIFF
--- a/examples/editor-libcosmic/src/text.rs
+++ b/examples/editor-libcosmic/src/text.rs
@@ -11,7 +11,7 @@ use cosmic::{
     iced_winit::renderer::BorderRadius,
     theme::Theme,
 };
-use cosmic_text::{Attrs, AttrsList, BufferLine, Metrics, SwashCache};
+use cosmic_text::{Attrs, AttrsList, BufferLine, HeightLimit, Metrics, SwashCache};
 use std::{cmp, sync::Mutex, time::Instant};
 
 pub struct Appearance {
@@ -49,6 +49,7 @@ impl Text {
 
         //TODO: make it possible to set attrs
         let mut line = BufferLine::new(string, AttrsList::new(Attrs::new()));
+        line.set_wrap(cosmic_text::Wrap::None);
 
         //TODO: do we have to immediately shape?
         line.shape(&crate::FONT_SYSTEM);
@@ -102,7 +103,7 @@ where
             limits.max().width as i32,
             self.line.wrap(),
             self.line.align(),
-            cosmic_text::Ellipsize::None,
+            cosmic_text::Ellipsize::End(HeightLimit::Default),
         );
 
         let mut width = 0;
@@ -168,7 +169,7 @@ where
             layout_w,
             self.line.wrap(),
             self.line.align(),
-            cosmic_text::Ellipsize::None,
+            cosmic_text::Ellipsize::End(HeightLimit::Default),
         );
 
         let mut cache = state.cache.lock().unwrap();
--- a/examples/editor-libcosmic/src/text.rs
+++ b/examples/editor-libcosmic/src/text.rs
@@ -102,6 +102,7 @@ where
             limits.max().width as i32,
             self.line.wrap(),
             self.line.align(),
+            cosmic_text::Ellipsize::None,
         );
 
         let mut width = 0;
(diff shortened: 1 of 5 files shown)


hojjatabdollahi:
Ellipsize::End for `Text` in `editor-libcosmic` example. 

image

hojjatabdollahi:
> The elipsis seems to be overlapping the switch (to it's right) here. That doesn't seem right...

That's actually the switch overlapping with the text. I wish I could set a minimum width for the UI elements in Iced!

Actually it would increase the complexity, especially because I'm thinking the multi-line ellipsization should happen in Buffer and not Bufferline to allow for ellipsizing multi-paragraph labels.

nicoburns:
> That's actually the switch overlapping with the text. I wish I could set a minimum width for the UI elements in Iced!

Ah, yeah. Iced's layout is... a bit lacking.

> Regarding the multi-line ellipsizing, it would increase the complexity, especially because I'm thinking the multi-line ellipsization should happen in Buffer and not Bufferline to allow for ellipsizing multi-paragraph labels.

Hmm... ok. Well, I'll defer to your expertise on this issue!

hojjatabdollahi:
> Hmm... ok. Well, I'll defer to your expertise on this issue!

Please don't! I'll take any help I can get. 

With multi-line ellipsize, I think GTK does it with a `set_line` function and I think I read somewhere that it only affects `Ellipsize::End` and won't do anything in other cases. But I'm not 100% sure.

geom3trik:
Hi, was just wondering if this is still being worked on? I was going to implement this in my UI framework but then noticed this PR to add it at the cosmic-text level.

hojjatabdollahi:
> Hi, was just wondering if this is still being worked on? 

Hi, 

I have been super busy irl. But this and a few RTL features are definitely somethings I will do. However, if you want to take over or help out, I would be happy. 

This was my plan:
I was thinking to change the visual line struct and add a property that describes if this visual line is ellipsized, hyphenated, wrapped, or none. That way instead of cosmic-text always adding an ellipsis glyph, the GUI framework can decide what it wants to do with visual lines that are ellipsized, if it wants to show ellipsis, or fade out the last few letters, or add some marker (like an arrow) that says the text is cut off.

airstrike:
Hi, just wondering if there's been any additional work on this or what would be needed to get it across the finish line

hojjatabdollahi:
I haven't had a chance to work on this and the codebase has changed a lot since when I started this PR, so I don't know how much more work is left to do.

Please feel free to build upon / copy / ignore this! I will close this PR as I won't be able to work on it for a few more months.

PR #86, opened 2023-02-28
Move fields from `Buffer` into a `BufferData` struct

geieredgar:
As noted by others in #39, keeping `Buffer`s around is cumbersome because it borrows `FontSystem`. This PR tries to solve that by making `Buffer` a wrapper around a new `BufferData` struct that contains all data previously owned by `Buffer`. 

`Buffer::from_data` and `Buffer::into_data` allow converting between `Buffer` and `BufferData`. 

Users of `Buffer` shouldn't need to change anything, because `Buffer` references are coerced to `BufferData` references via `Deref`/`DerefMut`.

Note: This is only a first step towards better ergonomics, because moving `BufferData` into `Buffer` can be itself pretty cumbersome and may require suboptimal code like using `Option<BufferData>`.
I envision the following steps for a follow-up:
1. Make `Buffer` borrow `BufferData` instead of owning it. The idea is that users would always store `BufferData` instead of `Buffer` and create a temporary `Buffer` when they need to work with it.
2. (Optional) If it is desired to keep an owning `Buffer`, we could instead rename `Buffer` to `OwnedBuffer`, and add something like this:
```rust
pub struct Buffer<'a> {
    font_system: &'a FontSystem,
    data: &'a mut BufferData,
}

pub trait BufferMethods {
    fn inner(&mut self) -> (&FontSystem, &mut BufferData);

    // Methods of the old Buffer would be moved here, with a default implementation 
    // that operates on the FontSystem and BufferData returned by inner
    ...
}

impl<'a> BufferMethods for Buffer<'a> {
    fn inner(&mut self) -> (&FontSystem, &mut BufferData) {
        (self.font_system, self.data)
    }
}  

impl<'a> BufferMethods for OwnedBuffer<'a> {
    fn inner(&mut self) -> (&FontSystem, &mut BufferData) {
        (self.font_system, &mut self.data)
    }
}
```

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -5,7 +5,10 @@ use alloc::{
     string::{String, ToString},
     vec::Vec,
 };
-use core::{cmp, fmt};
+use core::{
+    cmp, fmt,
+    ops::{Deref, DerefMut},
+};
 use unicode_segmentation::UnicodeSegmentation;
 
 #[cfg(feature = "swash")]
@@ -295,9 +298,8 @@ impl fmt::Display for Metrics {
     }
 }
 
-/// A buffer of text that is shaped and laid out
-pub struct Buffer<'a> {
-    font_system: &'a FontSystem,
+/// Buffer data for a text that is shaped and laid out
+pub struct BufferData {
     /// [BufferLine]s (or paragraphs) of text in the buffer
     pub lines: Vec<BufferLine>,
     metrics: Metrics,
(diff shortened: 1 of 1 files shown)


dhardy:
Is it possible to construct a `BufferData` without a `&FontSystem`? That would be the way I would want to use this.

KAS-text has an even lower level decomposition: [`Text`](https://docs.rs/kas-text/latest/kas_text/struct.Text.html) is a "text" + `Environment` (options) + `TextDisplay` (runs & glyphs without the source text), though in practice the only use I found for this is being generic over the text (`String`, &str` or some type of rich text providing formatting markers). The font library is completely excluded (it's a lazy static, and not accessed until the text is "prepared").

geieredgar:
Not with this PR alone, but I think a possible addition could be adding a flag to `BufferData` that signals `Buffer::from_data` to prepare the text.

geieredgar:
Closing this in favor of #97.

PR #87, opened 2023-02-28
Remove lifetime from `SwashCache`

geieredgar:
This removes the borrow of `FontSystem` from `SwashCache`, removing the need for a lifetime type parameter. Methods of `SwashCache` now take a `&FontSystem` instead, similar to #39. As the methods of `SwashCache` are few and, I believe,  will often not be used directly, the decrease in ergonomics should be tolerable.


geieredgar:
Conflicts are resolved.

jackpot51 (approved):

Issue #75, opened 2023-02-18
Allow access to &mut Database from FontSystem


notgull:
I'd like to be able to access the underlying `Database` in a `FontSystem` mutably, so I can easily add fonts without needing to convert to the underlying `Database` and back.

However, the `Database` is currently self-referentially borrowed for caching the font-ID-to-font relationship. It looks like this is due to the font lookup in `fontdb` being an `O(n)` process. RazrFalcon/fontdb#47 would fix that problem and allow us to solve this issue.

dhardy:
The approach in #89 was slow.

Alternative: allocate all fonts in a `Box<..>` so that they never move / deallocate, then cast references to `&'static`. Store the database in a global static with methods allowing new fonts be added only.

I'm not going to write a PR, but you could probably copy a lot of [my code](https://github.com/kas-gui/kas-text/blob/master/src/fonts/library.rs) since I already wrote a very similar library.

API: https://docs.rs/kas-text/latest/kas_text/fonts/struct.FontLibrary.html

Bonus feature: the library supports a [default font](https://docs.rs/kas-text/latest/kas_text/fonts/index.html#fontid-and-the-default-font) so that simple `Text` objects can be constructed without access to the font library.

PR #89, opened 2023-03-01
Make `FontSystem` not self-referencing and update `fontdb` and `rustybuzz`

geieredgar:
This makes the `std` implementation of `FontSystem` not self-referencing, which in turn allows defining a `db_mut` method for mutable access to the underlying `fontdb::Database` (requested in #75). 

The high level changes are:
- Adds a new `FontKey` struct that stores the `fontdb::ID` and (if the `swash` feature is enabled) the swash key data `(u32, swash::CacheKey)`.
  - A `FontKey` can be obtained via the new `FontSystem::get_font_key(id)`.
  - `FontKey`s are cached instead of caching `Font`s. 
- `FontSystem::get_font` now takes a `FontKey` instead of `fontdb::ID` and creates a `Font` on the fly, reusing the key data.
- Removes `FontMatches`.
   - `FontSystem::get_font_matches` now returns a `Arc<Vec<FontKey>>` instead of `Arc<FontMatches<'a>>`.
   - `FontFallbackIter` borrows `fontdb::Database` and creates the font on the fly using `Font::from_key(db, key)`.


notgull:
Just a heads up: before version 0.13.0, lookups in `fontdb` take linear time. As part of this PR, if you're making breaking changes anyways, I would bump `fontdb` to 0.13.0 so that lookups take `O(1)` time.

src/font/mod.rs
```diff
@@ -4,12 +4,17 @@ use core::ops::Deref;
...
-
 pub use self::system::*;
 mod system;
 
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+/// Identifies a [`Font`] in a [`FontSystem`]
+pub struct FontKey {
+    pub id: fontdb::ID,
+    #[cfg(feature = "swash")]
+    pub swash: (u32, swash::CacheKey),
```
CryZe, line 15: This potentially should be repr(transparent) if swash isn't active (I forgot if you are allowed to use transparent with multiple fields)

geieredgar: Done.

geieredgar:
I  fixed the duplicate `ttf-parser` entries by bumping `rustybuzz` to version `0.7`. Not sure I can do something about the failed license requirement.

geieredgar:
I've added `Zlib` to the list of allowed licenses. I think there is no harm in that.

jackpot51 (approved):

jackpot51:
@geieredgar I merged this since CI tests were passing but it appears to have drastically reduced the performance of the editor-test example, so I reverted. Please see if you can replicate this issue.

geieredgar:
Yes, I could replicate this issue. The reason is that `rustybuzz::Face::from_slice` takes a lot of time. I wrongly assumed it would behave similar to `ttf_parser::Face::parse` in terms of performance, which is intended to be cheap and therefore can be executed on the fly, but that isn't the case. I will look into trying to make `rustybuzz::Face::from_slice` not as expensive and then compare the performance of my approach with the old approach. Sorry for wasting your time!

PR #94, opened 2023-03-03
Update `fontdb` and `rustybuzz` dependency

geieredgar:
These are the isolated update commits from #89, which was reverted due to slow performance.

### DIFF
--- a/src/font/system/std.rs
+++ b/src/font/system/std.rs
@@ -75,8 +75,7 @@ impl FontSystem {
             let now = std::time::Instant::now();
 
             //TODO only do this on demand!
-            for i in 0..db.faces().len() {
-                let id = db.faces()[i].id;
+            for id in db.faces().map(|face| face.id).collect::<Vec<_>>() {
                 unsafe {
                     db.make_shared_face_data(id);
                 }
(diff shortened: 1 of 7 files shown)


src/font/system/std.rs
```diff
@@ -75,8 +75,7 @@ impl FontSystem {
             let now = std::time::Instant::now();
 
             //TODO only do this on demand!
-            for i in 0..db.faces().len() {
-                let id = db.faces()[i].id;
+            for id in db.faces().map(|face| face.id).collect::<Vec<_>>() {
```
MarijnS95, line 78: No need to collect into a temporary vec when you're only processing the `id`s one at a time and not storing this vec after? Same above.

geieredgar: `db.faces()` returns an iterator now, because of the switch to `slotmap`. So iterating over `faces()` borrows `db`, but we also need to borrow `db` mutably inside the loop for the `db.make_shared_face_data(id)` call.
I also don't like this, do you have a better idea to solve this? I am also working on a PR that will remove the need for the loop, so hopefully this will only be a temporary issue.

MarijnS95: Hmm, that is quite ugly. Funnily enough we have a PR open to get values as a slice instead of iterator from `slotmap`, but that won't solve the borrow problem. Will look forward to that PR then :)

geieredgar: Yeah, even I thought, after cherry-picking the commit from #89, "Do I really need this" and tried to remove it.

tosti007 (approved):

jackpot51 (approved):

PR #95, opened 2023-03-03
Make `FontSystem` not self-referencing, second try

geieredgar:
Like #89, this makes the `std` implementation of `FontSystem` not self-referencing, which in turn allows defining a `db_mut` method for mutable access to the underlying `fontdb::Database` (requested in #75). 

Unlike #89, this makes `Font` self-referencing. Fonts are cached like before, so the performance should stay the same. The only impact on performance should be that `FaceInfo` is cloned and stored inside the `Font` to remove the need to borrow from `fontdb::Database`.


geieredgar:
Closing this in favor of #97. The latest changes of #97 don't require cloning `FontFace` anymore.

PR #97, opened 2023-03-03
Remove lifetime from `Buffer` and `Mutex` from `FontSystem` 

### LINKED ISSUE #57: Loading fonts incrementally

### LINKED ISSUE #75: Allow access to &mut Database from FontSystem

### LINKED ISSUE #91: FontSystem::new() takes a significant amount of time
On my system (Xubuntu), running `FontSystem::new()` takes a significant amount of time. On release, it takes around a second to run. On debug, it takes up to ten seconds.

There should probably be some measures taken to make sure `FontSystem` runs quickly. However, it should also be noted somewhere that `FontSystem` takes a while to load.

### LINKED ISSUE #107: 96.8MB memory usage in rich-text example
Originally I found this issue in the advanced text branch for iced. My system reports that 96.8MB is used when running. It seems like a bit much to use that much memory for just displaying some text. Iced gave memory usage of around 100MB for the checkbox example on the advanced-text branch.

I ran dhat from valgrind for iced and it seems that some parts of rustybuzz appears to be the cause of 12MB of memory usage.

I am running on Linux on aarch64.

geieredgar:
This includes all changes from #95.

This makes the `Buffer` and `Editor` types not borrow `FontSystem`, changing their methods to take a `&mut FontSystem` when needed, like #39 does. A helper type `BorrowedWithFontSystem` is added (this was proposed here: https://github.com/pop-os/cosmic-text/pull/39#issuecomment-1445534417), which allows calling methods without having to pass a `FontSystem` for each call.

Making `FontSystem::get_font` take `&mut self` also allows us to lazily call `make_shared_face_data`.

This also adds an `FontSystem::db_mut` method, allowing mutable access to the `fontdb::Database`, which also clears the font match cache to make sure that newly added fonts can be matched. This would fix #75 and resolve #57. It also fixes #107.


notgull:
By the way, this wouldn't fix #91. There are still outstanding issues with font loading that still slow it down in certain cases.

geieredgar:
@notgull Could you check if this improves the load times? I've added a commit so it should only load the fonts that are really needed/used.

i509VCB:
> @notgull Could you check if this improves the load times? I've added a commit so it should only load the fonts that are really needed/used.

This does appear to improve the loading time on my system. You can notice `main` wait a moment to load the fonts. With this branch it does still take a tiny bit but it's generally done when Plasma's resize animations finish.

MarijnS95:
> Making `FontSystem::get_font` take `&mut self` also allows us to lazily call `make_shared_face_data`.

User question: I'm upgrading to `cosmic-text 0.10` in an internal crate, and have a loop of this form:

```rust
        for face in system.db().faces() {
            let font = system.get_font(face.id).unwrap();
            let font_ref = font.as_swash();
```

This won't compile because the `.faces()` iterator holds a borrow on `system`, while `get_font()` also wants to mutably borrow `system`. Is there a better solution than precomputing the list of IDs with `let face_ids = system.db().faces().map(|face| face.id).collect::<Vec<_>>();`?

PR #98, opened 2023-03-06
Fix word wrapping in `ShapeLine::layout`

geieredgar:
Playing around with the `editor-libcosmic` example I encountered a crash when using word wrapping, because the first word was larger than the line. This is caused by this line https://github.com/pop-os/cosmic-text/blob/1bc198fd6306923e797dcf022439faf44c343af0/src/shape.rs#L782,  because `i` is `0`.

~After fixing that I also noticed that if a word is larger than the line, it will cause an overflow. I think almost everywhere word wrapping implies wrapping at the glyph level, if the word does not fit the line, so I believe users will expect that when using `Wrap::Word`.~ 

~Therefore I changed the `ShapeLine::layout` method to wrap at the glyph level (Edit: see https://github.com/pop-os/cosmic-text/pull/98#issuecomment-1456340645 and other comments in this PR), if the wrapped word does not fit the new line~. 

I noticed some more issues with the current implementation:

- https://github.com/pop-os/cosmic-text/blob/1bc198fd6306923e797dcf022439faf44c343af0/src/shape.rs#L707 In the incongruent directions case this must be  `i + 1`.
- https://github.com/pop-os/cosmic-text/blob/1bc198fd6306923e797dcf022439faf44c343af0/src/shape.rs#L782 In both direction cases this ignores the fact that the current word could be the first in the line. However, it could be that this might work regardless, but that doesn't seem obvious.
- https://github.com/pop-os/cosmic-text/blob/1bc198fd6306923e797dcf022439faf44c343af0/src/shape.rs#L738 This should be `(i, 0)`.

~My implementation works by moving the `Wrap::Glyph` wrapping part behind the `Wrap::Word` case and executing it unless `continue` is used to skip it. Because the `Wrap::None` case is handled before the loop, we only have to deal with `Wrap::Word` and `Wrap::Glyph` so that should work.~

I've also renamed the local `word_size` variable to `word_width` for more consistency and because I think it's a more precise name.

I replaced the fetching of the previous word with a local `trailing_space_width` variable, which keeps track of the width of the last processed word in the current line. It is only `Some(_)` if we already processed a word in the current line and that word was blank.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -668,17 +668,23 @@ impl ShapeLine {
                 if self.rtl != span.level.is_rtl() {
                     // incongruent directions
                     let mut fitting_start = (span.words.len(), 0);
+                    // Keeps track of the width of the last word in the line, if it is blank
+                    let mut trailing_space_width = None;
                     for (i, word) in span.words.iter().enumerate().rev() {
-                        let word_size = font_size * word.x_advance;
-                        if fit_x - word_size >= 0. {
+                        let word_width = font_size * word.x_advance;
+                        if fit_x - word_width >= 0. {
                             // fits
-                            fit_x -= word_size;
-                            word_range_width += word_size;
+                            fit_x -= word_width;
+                            word_range_width += word_width;
                             if word.blank {
                                 number_of_blanks += 1;
+                                trailing_space_width = Some(word_width);
+                            } else {
+                                trailing_space_width = None;
                             }
                             continue;
-                        } else if wrap == Wrap::Glyph {
+                        }
+                        if wrap == Wrap::Glyph {
                             for (glyph_i, glyph) in word.glyphs.iter().enumerate().rev() {
                                 let glyph_size = font_size * glyph.x_advance;
                                 if fit_x - glyph_size >= 0. {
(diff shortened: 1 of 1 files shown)


nicoburns:
> I think almost everywhere word wrapping implies wrapping at the glyph level, if the word does not fit the line, so I believe users will expect that when using Wrap::Word.

CSS allows you to configure both options here (and the default is for single words that overflow to just overflow). I think we should provide both options. See https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

geieredgar:
Sounds reasonable, do you have a good name for the new `Wrap` variant?

geieredgar:
On second thought, we could also change the `Word` variant to `Word { overflow: bool } `.

hojjatabdollahi:
I fixed that `get` function in the Ellipsize branch, but hadn't got around to send a PR for it yet.

Regarding the wrapping, I was going to add it as a new wrapping mode called "Word-Glyph". GTK also has "None, Char, Word, Word-Char". I postponed this until we have hyphenation (in case we want to wrap large words, and not necessarily longer than the width words). There is now clear way (right now) to indicate to the user that the word is broken down now.

nicoburns:
I'd definitely prefer a flat enum rather than nesting a boolean is possible. Regarding naming: perhaps `PreferWord` or `WordWithGlyphFallback`. I actually think I like the latter most. It's verbose, but I think that can be a good thing.

geieredgar:
I've added the new variant and changed the code to only split at glyph level if it is used. I also changed the defaults to use the new variant. I only wasn't sure about the debug formatting of the variant.

hojjatabdollahi:
Scratch that. The issue with BiDi exists in the main branch too.

geieredgar:
I've reverted the defaults and removed the `Wrap::WordWithGlyphFallback` variant for now. I still left the changed order of `Glyph` and `Word` wrapping cases, I hope that is okay. I will open another PR that will revert these two last commits (or only the last commit, if `Word` should stay the default) and adds some tests, although I would need some guidance about what I should test and how to verify that I didn't break something.

geieredgar:
Restored the old order and squashed the commits again to make the PR easier to review.

hojjatabdollahi:
After doing more testing I noticed that when it comes to BiDi the algorithm to detect the trailing space is not correct. 

I think a better option is to just revert the commit 27d52a12 until we come up with a better solution for removing the trailing spaces that works better with BiDi.

geieredgar:
What exactly did you test? Can you confirm that these changes here are also not working correctly?

hojjatabdollahi:
Yes. 
Try opening `sample/farsi.txt` and keep changing the window size.
 
Some text in the BiDi line overflows or gets duplicated. In the LTR line sometimes the space before the last word disappears.

hojjatabdollahi:
@geieredgar  I had to open a new PR to fix the layouting issues that I mentioned. I tried to work off of your PR but I couldn't get the whitespaces in justified BiDi text to work properly. I'm sorry. I hope you don't mind.

geieredgar:
@hojjatabdollahi Of course I don't mind. It's great to hear you could fix those issues.

Closing this as superseded by #108.

PR #100, opened 2023-03-06
Unify `Attrs` and `AttrsOwned`

geieredgar:
Currently, there exist the borrowing `Attrs` and the owning `AttrsOwned`, which only differ in their usage of `Family` vs `FamilyOwned`. In theory, this allows passing `Attrs` without having to allocate. In practice, we have the following situation:
1. `FontSystem::get_font_matches` wants to prevent an allocation in the most-common case that a cached match already exists, but it can't (see https://github.com/pop-os/cosmic-text/blob/1bc198fd6306923e797dcf022439faf44c343af0/src/font/system/std.rs#L130 ), because we cannot implement `Borrow<Attrs>` for `AttrsOwned`.  
2. We need to store `AttrsOwned` in `AttrsList`, so it is almost impossible to not have an `Attrs` turn into an `AtttsOwned` somewhere down the line.

This PR removes the old `Attrs`, renames `AttrsOwned` to `Attrs` and adds an `AttrsBuilder`, which has the same helper methods as the old `Attrs`. This PR also includes several changes to prevent allocations in many cases:

- `FamilyOwned` uses a `Cow<'static, str>` instead of `String`, allowing the use of string literals without allocating.
- `AttrsBuilder::family` takes an `impl OwnedOrStaticFamily`, which is only implemented for `FamilyOwned` or `Family<'static>`, so users need to explicity use `FamilyOwned::new` if they want to use a non static `Family`.
- `Attrs` implements `AsRef<Attrs>` and `From<&Attrs>`, allowing methods that might need to store the `Attrs` to accept both values and references of `Attrs`, while at the same time preventing an extra allocation when a value is passed.  
- As a side effect of `Attrs` now not being `Copy`, users have to explicitly use `clone`, which makes potential allocations more obvious.

This PR also changes `FontSystem::get_font_matches` to take an `impl AsRef<Attrs> + Into<Attrs>` parameter, which it only clones if (a) the passed argument is a reference and (b) the cache does not contain an equivalent `Attrs`.


jackpot51:
This now has conflicts after merging #97

geieredgar:
The conflicts are resolved, but the CI fails right now, caused by duplicate `syn` dependency, unrelated to my changes.

Imberflur:
FWIW point 1 can probably be addressed via using `hashbrown` with [HashMap::entry_ref](https://docs.rs/hashbrown/0.14.0/hashbrown/struct.HashMap.html#method.entry_ref) + [Equivalent](https://docs.rs/hashbrown/0.14.0/hashbrown/trait.Equivalent.html).

genusistimelord (commented):
I think the changes here seem reasonable as it will help simplify the overall usage of Attr's. 

My only Concern would be how will this affect the overall performance since we are moving towards clones more than copy. @jackpot51 what do you think?

PR #101, opened 2023-03-06
Extract color overrides from `attrs_list` into new `color_spans` field

geieredgar:
This is a follow up for #100. It removes the `color_opt` field from `Attrs` and stores the information in the new `color_spans` field of `BufferLine` instead. This allows it to be mutated without having to do relayouts.


genusistimelord:
this looks very promising, However You probably should move Metadata in with the Color as these are ones that don't affect the Layout.

PR #102, opened 2023-03-06
Replace use of `str::lines` with `unicode-bidi`

### LINKED ISSUE #74: assertion failed when source text contains \r character
MRE:
```rust
// cosmic-text = "0.7.0"
use cosmic_text::{Attrs, Buffer, FontSystem, Metrics};

fn main() {
    let font_system = FontSystem::new();
    let metrics = Metrics::new(14, 20);
    let mut buffer = Buffer::new(&font_system, metrics);
    buffer.set_size(80, 25);
    let attrs = Attrs::new();
    let message = "A\rB"; // Problem!
    buffer.set_text(message, attrs);
}
```

Discovered while trying to develop an X11 application- the Enter key sent a "\r" instead of a "\n", which my prototype application (not using `Editor`) simply appended to the end of the rendered buffer, causing a crash when the next character was appended.

The cause seems to be that `\r` is treated as a paragraph separator by `unicode-bidi`, but not by the `lines` iterator from std uses to do paragraph separator in `buffer.set_text()`, so when the paragraph is sent to be shaped it's split into two paragraphs and triggers the assert.

geieredgar:
This should fix #74.

### DIFF
--- a/examples/editor-test/Cargo.toml
+++ b/examples/editor-test/Cargo.toml
@@ -13,3 +13,8 @@ fontdb = "0.13"
 log = "0.4"
 orbclient = "0.3.35"
 unicode-segmentation = "1.7"
+
+[dependencies.unicode-bidi]
+version = "0.3.8"
+default-features = false
+features = ["hardcoded-data"]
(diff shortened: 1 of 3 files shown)


geieredgar:
I didn't consider that the paragraph ranges, unlike `str::lines`, contain the paragraph delimiter. This is fixed now.

PR #108, opened 2023-03-13
Fix a few bugs in layout causing the text to overflow or disappear

hojjatabdollahi:
While testing #98 I noticed a few layouting issues. These issues are in the main branch too. 

One issue was in BiDi text, sometimes there was an overflow like the word "Left" in this picture:

Screenshot from 2023-03-12 21-47-01

This issue sometimes causes the last word to repeat or to disappear.

Another issue was in the LTR-only lines (or RTL-only), the space before the last word would disappear if the linewidth was a specific size. Like in the image below:

Screenshot from 2023-03-12 21-52-05

In this PR I fix these issues by changing the layout algorithm a little. Previously we used to create `word_ranges` and later put them together in a `visual_line` but now the `visual_line` is created while iterating over the words. This simplifies the algorithm and fixes the issue mentioned above.

This PR supersedes #98. I tried to apply these changes to @geieredgar 's PR, but the algorithm they used for detecting trailing spaces would not work in specific situations in BiDi. So, I decided to open a new PR.


src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
-                            if word.blank && number_of_blanks > 0 {
-                                // current word causing a wrap is a space so we ignore it
-                                number_of_blanks -= 1;
-                            } else if let Some(previous_word) = span.words.get(i - 1) {
-                                // Current word causing a wrap is not whitespace, so we ignore the
-                                // previous word if it's a whitespace
-                                if previous_word.blank {
-                                    number_of_blanks -= 1;
-                                    prev_word_width = Some(previous_word.x_advance * font_size);
+                            let mut trialing_space_width = None;
```
geieredgar, line 724: This contains a typo: It should be renamed to `trailing_space_width`.

src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
-                                    fit_x = line_width - glyph_size;
-                                    word_range_width = glyph_size;
+                                    fit_x = line_width - glyph_width;
+                                    word_range_width = glyph_width;
                                     fitting_start = (i, glyph_i);
                                 }
                             }
                         } else {
                             // Wrap::Word
                             let mut prev_word_width = None;
```
geieredgar, line 818: I think for better consistency, we should rename this to `trailing_space_width` too.

src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
-                                number_of_blanks -= 1;
-                            } else if let Some(previous_word) = span.words.get(i - 1) {
-                                // Current word causing a wrap is not whitespace, so we ignore the
-                                // previous word if it's a whitespace
-                                if previous_word.blank {
-                                    number_of_blanks -= 1;
-                                    prev_word_width = Some(previous_word.x_advance * font_size);
+                            let mut trialing_space_width = None;
+                            if i > 0 {
+                                if let Some(previous_word) = span.words.get(i - 1) {
```
geieredgar, line 726: Hmm, this is something I changed to `i + 1` in my PR. I'm just curious, so can you explain to me why this is correct? Or should the variable be renamed to `next_word`?

src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
                                 }
                             }
-                            if let Some(width) = prev_word_width {
-                                word_ranges.push((
-                                    (i, 0),
+                            if let Some(width) = trialing_space_width {
+                                self.add_to_visual_line(
+                                    &mut current_visual_line,
+                                    span_index,
+                                    (i + 2, 0),
```
geieredgar, line 740: Sorry for asking so many questions, but I'm just curious, since this differs so much from the congruent direction case. Why do we use `i + 2` here, but `i` in the other direction?

src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
                                     word_range_width - width,
                                     number_of_blanks,
-                                ));
+                                );
                             } else {
-                                word_ranges.push((
+                                self.add_to_visual_line(
+                                    &mut current_visual_line,
+                                    span_index,
                                     (i + 1, 0),
```
geieredgar, line 749: Same as above, why is it `i + 1` here, but just `i` in the other direction?

src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
+                                );
                             }
+                            visual_lines.push(current_visual_line);
+                            current_visual_line = VisualLine::default();
+
                             number_of_blanks = 0;
                             if word.blank {
                                 fit_x = line_width;
                                 word_range_width = 0.;
                                 fitting_start = (i + 1, 0);
```
geieredgar, line 762: I also don't understand why this is `i + 1` instead of `i - 1` if we are going in the opposite direction.

src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
                             } else {
-                                fit_x = line_width - word_size;
-                                word_range_width = word_size;
+                                fit_x = line_width - word_width;
+                                word_range_width = word_width;
                                 fitting_start = (i + 1, 0);
```
geieredgar, line 766: Same question: Why is this `i + 1` here, but just `i` in the congruent direction case.

src/shape.rs
```diff
@@ -660,237 +677,214 @@ impl ShapeLine {
...
-                                    prev_word_width = Some(previous_word.x_advance * font_size);
+                                // number_of_blanks = number_of_blanks.saturating_sub(1);
+                            } else if i > 0 {
+                                if let Some(previous_word) = span.words.get(i - 1) {
+                                    // Current word causing a wrap is not whitespace, so we ignore the
+                                    // previous word if it's a whitespace
+                                    if previous_word.blank {
+                                        prev_word_width = Some(previous_word.x_advance * font_size);
+                                        number_of_blanks = number_of_blanks.saturating_sub(1);
+                                    }
```
geieredgar, line 829: Why does it matter here that the current word is blank, but not in the incongruent direction case?

I also think this should be collapsed into one a single `if !word.blank && i > 0`.

geieredgar (changes requested):
I have some smaller nitpicks and some questions about why the incongruent and congruent direction cases differ so much.

src/shape.rs
hojjatabdollahi: Hopefully one day we can combine normal if and if-let together. Until then I think this is fine.

src/shape.rs
hojjatabdollahi: Fixed in [7fa51c6](https://github.com/pop-os/cosmic-text/pull/108/commits/7fa51c6404f7f286a4c23abc79640bc0e63a5138)

src/shape.rs
hojjatabdollahi: Fixed in [7fa51c6](https://github.com/pop-os/cosmic-text/pull/108/commits/7fa51c6404f7f286a4c23abc79640bc0e63a5138)

src/shape.rs
```diff
@@ -609,6 +615,24 @@ impl ShapeLine {
...
+        number_of_blanks: u32,
+    ) {
+        if end == start {
+            return;
+        }
+
+        vl.ranges.push((span_index, start, end));
+        vl.w += width;
+        vl.spaces += number_of_blanks;
+    }
```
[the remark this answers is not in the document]
hojjatabdollahi: Moved it into the layout function in [7fa51c6](https://github.com/pop-os/cosmic-text/pull/108/commits/7fa51c6404f7f286a4c23abc79640bc0e63a5138)

src/shape.rs
hojjatabdollahi: Here is how the algorithm currently works:

- When we are shaping a some text. Depending on the directionality of the first character we decide the direction of the whole paragraph. So if the paragraph starts with an RTL character and the rest is all LTR, we decide that that whole paragraph is RTL.

- We then create `spans` of text. Each span is created based on the text direction. So, if the whole paragraph is LTR, the whole paragraph is only one span. But if you have BiDi text, which means a mix of LTR and RTL, then the paragraph is chopped up in smaller spans. **Each span always has only one direction. Either LTR or RTL.**

- When we are shaping an LTR span inside of an RTL paragraph (or the other way around) we store the words for that span in a reverse order. It happens [here](https://github.com/pop-os/cosmic-text/blob/c4a8d521f60ad35f95b083256ae993de2537d3a0/src/shape.rs#L419).

- Now, in the `ShapeLine::layout()` function, we iterate over these spans and find out what fits inside the line_width. If the direction of the paragraph is the same as the direction of the span, this process is pretty straight forward. But for the cases that the directions differ we have to iterate through them using `.rev()`. 

- In the layout function, we find the ranges of words inside of a span that fits inside a line. So, if we have a long span of words. It will take multiple lines. So, the way we are saving it is `(span_index, start, end)`. Each `VisualLine` has a `Vec<(span_index, start, end)>`. 

- Later we create `LayoutLine` using the ranges of words inside the `VisualLine`. `LayoutLine` contains the information about the glyphs and their positions.

As an example Let's say we have a paragraph that starts with a very long English text (LTR) and has some Farsi in it too (RTL). Let's say our desired output should look like this:
```
LLL LL LLLL LLLL LL LLL LL
LLLL LLLLL LLLL R RRR
RR RRRR LLLL LLL.
```
Now we have 3 spans in this paragraph:
```
span 0: LLL LL LLLL LLLL LL LLL LL LLLL LLLLL LLLL 
span 1: RR RRRR  R RRR 
span 2: LLLL LLL.
```
The first visual line would be pretty easy: 
```
Visual Line 0: [(0, 0, 7)]
```
It is saying span 0, from the first word till the 7th word (index 0 to 7, not inclusive on 7).

The second visual line is different! It starts with the rest of the first span, but when it gets to the RTL text the indicies get a little trickier.

If we didn't have a line break the RTL text would look like this: `RR RRRR  R RRR`, but now that we have a line break the two right-most words would fit but the two left-most words would be placed in the next line.

So the visual line would look like:
```
Visual Line 1: [(0,7,10), (1, 2, 4)]
```
and the last visual line would be:
```
Visual Line 2: [(1, 0, 2), (2, 0, 2)]
```



Now, when we are iterating over words in the case that the paragraph direction matches the span direction, the `i` would mean the _next_ word (visually). But in case of incongruent directions, the `i` means _previous_ word (visually).
Since we also are going in reverse direction, if we find a word that doesn't fit we store `(span_index, i+1, fitting_start)` as the range of words that would fit. Since we are moving backwards `fitting_start` is bigger than `i`.

Now if the previous word was a blank, then we ignore that and save `i+2` as the start of that word range. 

**Notice that the start of a word range is inclusive.**

If the direction of the span and the paragraph are the same and if we find a word that doesn't fit we store `(span_index, fitting_start, i)`. If the previous word was a blank we store `(span_index, fitting_start, i-1)`. Since in this case we are iterating in the correct direction (and not reversed) then ` i > fitting_start`.

**Notice that the end of the word range is not inclusive.**

Ok. This is the basic algorithm. However, since we also want to be able to wrap in a character level, we need to also say from which glyph to which glyph it would fit inside a line.

So our Visual Lines are actually `(span_index, (starting_word_index, starting_word_glyph_index), (ending_word_index, ending_word_glyph_index)`. Also note that it is possible to have a long word that is broken over multiple lines.

So if you have a word range that says `(0, (0,0), (0,10))`. That means the algorithm could only fit 10 glyphs from the first word of the first span.

If the word range says `(0, (0,0), (4,0))`, it means we were able to fit the first 4 words of the first span.

If the word range says `(0, (2,1), (3,5))`, it means from the second glyph of the third word, until the 5th glyph of the forth word.

I hope this helped.

src/shape.rs
```diff
@@ -660,237 +676,215 @@ impl ShapeLine {
...
+                                );
                             }
+                            visual_lines.push(current_visual_line);
+                            current_visual_line = VisualLine::default();
+
                             number_of_blanks = 0;
                             if word.blank {
                                 fit_x = line_width;
                                 word_range_width = 0.;
                                 fitting_start = (i + 1, 0);
```
geieredgar, line 761:
```suggestion
                                fitting_start = (i, 0);
```
In the congruent case, if the word is blank we don't include it in the next visual line by setting `fitting_start = (i + 1, 0)`, because the start index is inclusive. If the same should happen in the incongruent case, `fitting_start` should be set to `(i, 0)` because the end index is exclusive. With `i + 1` the blank word would still be included in the next visual line.

As a side note, does it make sense to rename `fitting_start` to `fitting_end` in the incongruent case because it will always be used as an end index?

src/shape.rs
```diff
@@ -660,237 +676,215 @@ impl ShapeLine {
...
                             }
-                            if let Some(width) = prev_word_width {
-                                word_ranges.push((
+                            if let Some(width) = trailing_space_width {
+                                add_to_visual_line(
+                                    &mut current_visual_line,
+                                    span_index,
                                     fitting_start,
-                                    (i - 1, 0),
+                                    (i, 0),
```
geieredgar, line 837:
```suggestion
                                    (i - 1, 0),
```
This handles the case when the last word that fits in the line was blank, right? Previously, the `i - 1` made sure to skip the blank word. In the incongruent case the blank word is skipped by using `i + 2` as start index, so it makes sense that it should also be skipped in the congruent case by using `i - 1` as end index. Or is there another reason this was changed?

src/shape.rs
geieredgar:
```suggestion
                                if let Some(previous_word) = span.words.get(i + 1) {
```
I still believe this should be `i + 1`, because we are iterating over descending `i` and the last word added to the line had the index `i + 1`, not `i - 1`.

geieredgar (changes requested):

src/shape.rs
geieredgar: Also, I think the `i > 0` check in the incongruent case is wrong. In the congruent case we don't have a previous word if `i == 0`, but this is not true in the incongruent direction. There we don't have a "previous" word if `i + 1 == span.words.len()`, but we don't need to take special care of that because `get` returns `None` if the index is out of bounds.

src/shape.rs
geieredgar: I agree, but I was only talking about the empty `if word.blank` case. So that 
```rust
if word.blank {
} else if i > 0 {
   ...
}
```
becomes
```rust
if !word.blank && i > 0 {
    ...
}
```
I also now think the `word.blank` check is unnecessary (we don't do it in the incongruent direction case), so it should probably be just
 ```rust
if i > 0 {
    ...
}
```

geieredgar:
@hojjatabdollahi Thank you again for your explanation, it helped me to clear up some misconceptions I had. Unfortunately; I cannot mark conversations as resolved, so it would be nice if you could do that to remove the clutter from already addressed comments.

src/shape.rs
geieredgar: This is correct, as pointed out in https://github.com/pop-os/cosmic-text/pull/108#discussion_r1134197269.

src/shape.rs
geieredgar: My confusion was cleared up here: https://github.com/pop-os/cosmic-text/pull/108#discussion_r1134197269.

src/shape.rs
hojjatabdollahi: Thanks for catching this. I forgot to remove the extra `if word.blank`. I removed it in the other direction. But forgot this one.

src/shape.rs
hojjatabdollahi: Good catch. It should be `i + 1`.

src/shape.rs
hojjatabdollahi: Since I don't remember why I changed it and your reasoning is sound I'll revert it.

src/shape.rs
hojjatabdollahi: it should be `i`. I've been trying to fix the trailing space that falls between the transition between LTR+RTL texts. Some of these changes were for that. But I haven't figured it out yet.
Like this space:

Screenshot from 2023-03-13 12-59-35

hojjatabdollahi:
@geieredgar Thanks for your review. I tested your suggested changes and I didn't see any regression, it fixed a potential bug and it also made congruent and incongruent more consistent. Thank you very much.

geieredgar (approved):
This looks good, I have no open questions, LGTM.

jackpot51:
All tests passed, looks good to me

PR #109, opened 2023-03-14
Allow discarding of unused fonts

geieredgar:
With these changes, `BufferLine` keeps track of the fonts used when it is shaped. It also adds a `FontSystem::discard_unused_fonts` method that removes those fonts that aren't references outside the font cache. 

I initially tried to implement this using `Weak`, to remove the need for the user to call `discard_unused_fonts`, but this had several issues:
- Like #89, the `editor-test` regressed due to many Font recreations, despite trying to hold onto the `Arc<Font>`s for as long as reasonably possible.
- Dropping `Font` is not enough for a true cleanup, because we would want to call `make_face_data_unshared` to free the memory mapping of unused fonts. This requires mutable access to the font database, which we cannot get during `Font::drop`.

I think having to call `discard_unused_fonts` is not the worst thing, we could also add methods to estimate the memory usage of cached fonts which could help users to decide when to discard unused fonts to free memory.

### DIFF
--- a/src/buffer_line.rs
+++ b/src/buffer_line.rs
@@ -1,7 +1,8 @@
+use alloc::sync::Arc;
 #[cfg(not(feature = "std"))]
 use alloc::{string::String, vec::Vec};
 
-use crate::{Align, AttrsList, FontSystem, LayoutLine, ShapeLine, Wrap};
+use crate::{Align, AttrsList, Font, FontSystem, LayoutLine, ShapeLine, Wrap};
 
 /// A line (or paragraph) of text that is shaped and laid out
 pub struct BufferLine {
@@ -12,6 +13,7 @@ pub struct BufferLine {
     align: Option<Align>,
     shape_opt: Option<ShapeLine>,
     layout_opt: Option<Vec<LayoutLine>>,
+    used_fonts: Vec<Arc<Font>>,
 }
 
 impl BufferLine {
@@ -26,6 +28,7 @@ impl BufferLine {
             align: None,
             shape_opt: None,
             layout_opt: None,
+            used_fonts: Vec::new(),
         }
     }
 
@@ -152,6 +155,7 @@ impl BufferLine {
     /// Reset shaping and layout information
     //TODO: make this private
     pub fn reset(&mut self) {
+        self.used_fonts.clear();
         self.shape_opt = None;
         self.layout_opt = None;
     }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -35,6 +35,7 @@ no_std = [
 std = [
   "fontdb/memmap",
   "fontdb/std",
+  "ouroboros/std",
```
i509VCB, line 38: If ouroboros is becoming optional based on features, this should be made the following instead to ensure enabling std does not needlessly pull in ouroboros

```suggestion
  "ouroboros?/std",
```

If that's not intended, then it should probably be noted since I am a bit confused if that is the case.

geieredgar: Ouroboros is not becoming optional, but rather becoming mandatory. Previously, ouroboros was only used for `FontSystem` if the `std` feature was used. Now it will be always used for `Font`. Note that I didn't change the ouroboros entry to be optional, I just disabled the default-features (`std`) and enabled it only if the `std` feature is used.

The current main is wrong with not disabling the `std` feature of ouroboros, but it didn't lead to issues because the ouroboros macro was only used when the `std` feature was active. I hope this makes sense.

i509VCB: Oh I guess I misread default-features as a optional...

genusistimelord:
this would be nice to have. I do think we could fix a lot of the referencing issues if we changed the Font system itself to give each Font a generational ID or just plain Key that can be used to link back to the Fonts. This way there is no need to Referencing Fonts. 

I think the API for this needs to be a bit different. Probably would go with a Collector and then a External font unloader. As you could have many Bufferlines through out the program that all use different fonts. you don't want to accidentally delete the fonts for them, so collecting them all together makes a lot more sense for this aspect.  Also If you redo this to use a Special Font ID that links to the Font in the Font system you could essentially use a HashSet to store which fonts got used. You could go even farther and set the Font usage inside the Font system itself and not have bufferline handle any of that.

dhardy:
*Is* there much utility to this?

It's something I never implemented in KAS-text for two reasons:

- If fonts are never unloaded, you can cast a `&Font` to a `&'static Font` and any once-valid font identifier is known to always be valid
- If a font is used once, there's a good chance it will be used again in the future. Even if not, the "lost memory" is not usually significant.

genusistimelord:
I think this is mostly Great for applications where you don't want all the Extra preloaded fonts to Exist in memory since you only use maybe one or two of them. OR you might be doing something like Microsoft Word and do not want to keep fonts you are no longer using loaded but can quickly reload them later when needed again.

PR #110, opened 2023-03-14
Use doc_auto_cfg for docs generation

i509VCB:
This is done when generating docs on docs.rs and locally with `RUSTDOCFLAGS="--cfg=docsrs" cargo +nightly doc --all-features`

There seems to be some bugs with the labels however:

There does seem to be a rustdoc bug when generating the labels as `SwashImage` fails to get a feature label automatically assigned to the type, but the functions have the label.

Also `FontSystem` seems to get an `std` label even though it is available `no_std`. I imagine that could be fixed by reworking the reexports.

Screenshot 2023-03-13 at 21-02-48 cosmic_text - Rust

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -12,8 +12,16 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+    # Right now the doc_cfg and auto_doc_cfg features require nightly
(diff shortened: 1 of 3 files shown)


i509VCB:
Going through dead pull requests, probably won't get to finishing this one due to other priorities.

PR #111, opened 2023-03-14
Warn on clippy::std_instead_of_core and other related lints.

i509VCB:
Primarily to let users know that some code might not compile if no_std mode is enabled.

This may be unnessecary for the `std` module of FontSystem, but it can make copying code between the two easier if needed.

### DIFF
--- a/src/font/system/std.rs
+++ b/src/font/system/std.rs
@@ -2,8 +2,9 @@
 
 use std::{
     collections::HashMap,
-    sync::{Arc, Mutex},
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jackpot51:
This now has conflicts after merging #97

i509VCB:
Going through dead pull requests, probably won't get to finishing this one due to other priorities (and I would need to start it again).

PR #118, opened 2023-03-28
Use checked_sub to fix panic on window resize.

13r0ck:
bug link: https://github.com/iced-rs/iced/issues/1773


13r0ck:
It seems there is no way to fix the `cargo-deny` failure currently? Ouroboros just depends on syn 1.0 ...

13r0ck:
Rebased on #119 to fix build

jackpot51 (approved):

PR #119, opened 2023-03-29
Fix `FontSystem::new_with_locale_and_db` Signature

CryZe:
It turns out that the `no_std` version of `FontSystem::new_with_locale_and_db` doesn't match the `std` version. This fixes that.

### DIFF
--- a/examples/editor-libcosmic/src/text_box.rs
+++ b/examples/editor-libcosmic/src/text_box.rs
@@ -10,7 +10,7 @@ use cosmic::{
         mouse::{self, Button, Event as MouseEvent, ScrollDelta},
         renderer,
         widget::{self, tree, Widget},
-        Padding, {Color, Element, Length, Point, Rectangle, Shell, Size},
+        Color, Element, Length, Padding, Point, Rectangle, Shell, Size,
     },
     theme::{Theme, ThemeType},
 };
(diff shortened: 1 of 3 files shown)


examples/editor-libcosmic/src/text_box.rs
```diff
@@ -227,7 +227,9 @@ where
 
         // Scale metrics
         let metrics = editor.buffer().metrics();
-        editor.buffer_mut().set_metrics(metrics.scale(style.scale_factor as f32));
+        editor
+            .buffer_mut()
+            .set_metrics(metrics.scale(style.scale_factor as f32));
```
CryZe, line 232: This was required to make the CI not fail.

CryZe:
oof, the syn problem is not going to be resolvable (easily at least)

CryZe:
@jackpot51 Should I deactivate the check in the deny.toml in the meantime? The ecosystem will have to go through this transition from syn 1 to syn 2 anyway. There isn't much cosmic-text can do about it.

jackpot51:
> @jackpot51 Should I deactivate the check in the deny.toml in the meantime? The ecosystem will have to go through this transition from syn 1 to syn 2 anyway. There isn't much cosmic-text can do about it.

Sure

jackpot51 (approved):

PR #122, opened 2023-04-21
`Shaping` strategy selection

hecrj:
This PR introduces a `Shaping` enum that allows users to control the shaping strategy used.

A new `Basic` shaping strategy is also added which naively positions glyphs and skips `rustybuzz` completely.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -21,6 +21,7 @@ pub enum Shaping {
     ///
     /// You should use this strategy when you have complete control of the text
     /// and the font you are displaying in your application.
+    #[cfg(feature = "swash")]
     Basic,
     /// Advanced text shaping and font fallback.
     ///
@@ -41,6 +42,7 @@ impl Shaping {
         span_rtl: bool,
     ) -> Vec<ShapeGlyph> {
         match self {
+            #[cfg(feature = "swash")]
             Self::Basic => shape_skip(font_system, line, attrs_list, start_run, end_run),
             Self::Advanced => {
                 shape_run(font_system, line, attrs_list, start_run, end_run, span_rtl)
@@ -251,6 +253,7 @@ fn shape_run(
     glyphs
 }
 
+#[cfg(feature = "swash")]
 fn shape_skip(
     font_system: &mut FontSystem,
     line: &str,
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -11,9 +11,22 @@ use unicode_segmentation::UnicodeSegmentation;
 use crate::fallback::FontFallbackIter;
 use crate::{Align, AttrsList, CacheKey, Color, Font, FontSystem, LayoutGlyph, LayoutLine, Wrap};
 
+/// The shaping strategy of some text.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Shaping {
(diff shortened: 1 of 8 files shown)


src/shape.rs
```diff
@@ -11,6 +11,46 @@ use unicode_segmentation::UnicodeSegmentation;
 use crate::fallback::FontFallbackIter;
 use crate::{Align, AttrsList, CacheKey, Color, Font, FontSystem, LayoutGlyph, LayoutLine, Wrap};
 
+/// The shaping strategy of some text.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Shaping {
```
nicoburns, line 16: Suggestion: I feel like this ought to be called `ShapingStrategy` or `ShapingAlgorithm` or similar rather than just `Shaping`.

src/shape.rs
```diff
@@ -11,6 +11,46 @@ use unicode_segmentation::UnicodeSegmentation;
...
+    /// Basic shaping with no font fallback.
+    ///
+    /// This shaping strategy is very cheap, but it will not display complex
+    /// scripts properly nor try to find missing glyphs in your system fonts.
+    ///
+    /// You should use this strategy when you have complete control of the text
+    /// and the font you are displaying in your application.
+    #[cfg(feature = "swash")]
+    Basic,
```
nicoburns, line 25: It would be good if this gave some idea of which fonts, or in which circumstances this algorithm is likely to work.

src/shape.rs
```diff
@@ -213,6 +253,50 @@ fn shape_run(
...
+) -> Vec<ShapeGlyph> {
+    let attrs = attrs_list.get_span(start_run);
+    let fonts = font_system.get_font_matches(attrs);
+
+    let default_families = [&attrs.family];
+    let mut font_iter = FontFallbackIter::new(font_system, &fonts, &default_families, Vec::new());
+
+    let font = font_iter.next().expect("no default font found");
+    let font_id = font.id();
+    let font = font.as_swash();
```
nicoburns, line 272: Does it make sense for font fallback behaviour to be controlled by the same `Shaping` parameter? Would it make sense to expose this as a separate configuration option?

src/shape.rs
hecrj: I think "basic shaping" and "advanced shaping" do make sense in a text library without any additional context.

src/shape.rs
hecrj: It may. It's probably better to wait until we have a particular use case in mind, however.

I think the enum could be extended to accept a `fn` pointer at some point.

src/shape.rs
genusistimelord: I agree with hecj here that there really is no need to add more context to it if its just determining from two types of shaping. However If we were to further expand this ShapingStrategy would make the most logical sense here. This would allow use to go outside the Basic or Advanced scope and allow use to implement shaping strategy's from other libraries.

src/shape.rs
genusistimelord: I think it makes sense if we are to limit which Fonts the shaping could be used with. But this would also need to be Documented in the enum to give a clear idea of why this font or that font cant be loaded with this shaping algorithm.

jackpot51 (approved):

PR #127, opened 2023-05-12
Allow creating a `Buffer` with no `FontSystem`

tigregalis:
# Changes

This introduces a new constructor `Buffer::new_empty` which has zero `BufferLine`s and is unshaped.

`Buffer::new` calls `Buffer::new_empty`.

# Context

https://github.com/pop-os/cosmic-text/pull/97 untangled the `Buffer` from the `FontSystem`, but didn't remove the need to pass a `FontSystem` into `Buffer::new`.

It is useful to be able to initialise a `Buffer` when you don't have a `FontSystem` in scope, for example if you want to `impl Default` for some type that contains a `Buffer`, since it is otherwise constructable without a `FontSystem`:

```rust
let buffer = Buffer::new_empty(Metrics::new(MY_DEFAULT_FONT_SIZE, MY_DEFAULT_LINE_HEIGHT));
```

# Future work

Any of the examples where `Buffer::new` is called and then immediately overwritten through `Buffer::set_text` or overwriting the `BufferLine`s could be rewritten in terms of `Buffer::new_empty` but I have not done this.


jackpot51 (dismissed):

jackpot51:
This now has conflicts after #122

tigregalis:
I've resolved the conflict.

Issue #46, opened 2022-11-23
[Examples] Issues running examples


saona-raimundo:
Hi! First of all, thanks for the crate!

I was trying to run the examples but faced some issues and I was wondering 
Would accept PRs to improve them?

The issues I faced are the following. Some of them are from the "first-user perspective", I just want to help.
1. I was not sure what each example was about. There was no README or module documentation on the source code.
2. The `terminal` example uses `termion`, which does not compile on Windows... So I could not run it.
3. The `editor-orbclient` example compiles, but panic with `[2022-11-23T10:09:18Z ERROR editor_orbclient] failed to load "": The system cannot find the path specified.`
4. The `editor-libcosmic` example failed to build because `xdg::BaseDirectories` was not found.

(I am on Windows running `rustc 1.64.0`)
Some preliminary thoughts about these issues are the following.
1. The `terminal` example uses `termion` modules `color` and `cursor`. A more platform-friendly crate with these features may be [`console`](https://crates.io/crates/console). It also ranks higher in https://lib.rs/std so I thought the change would be an improvement.

I have not checked the other examples in detail yet.

CeleritasCelery:
I am also having issues running the example on MacOS. the `terminal` one was the only one that worked.

- `editor-orbclient` failed with `[2023-02-22T21:11:20Z ERROR editor_orbclient] failed to load "": No such file or directory (os error 2)`
- `editor-libcosmic` failed because the wayland dependency failed to compile (looks to be linux only).
- `editor-test` failed with 
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Cursor { line: 0, index: 1, affinity: Before }`,
 right: `Cursor { line: 0, index: 1, affinity: After }`', examples/editor-test/src/main.rs:104:17
```

PR #129, opened 2023-05-12
Update terminal example using `colored`

tigregalis:
The terminal example didn't work for me on Windows because `termion` is not supported on Windows.

- Removed the `termion` dependency
- Added `colored` as a dependency
- Instead of terminal control sequences, I emulate a canvas, write the pixels to a string, and print that to the terminal

I also added the ability to pass text as an argument.
image

Fixes part of #46

### DIFF
--- a/examples/terminal/src/main.rs
+++ b/examples/terminal/src/main.rs
@@ -1,8 +1,11 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 
+//! Run this example with `cargo run --package terminal`
+//! or `cargo run --package terminal -- "my own text"`
+
+use colored::Colorize;
 use cosmic_text::{Attrs, Buffer, Color, FontSystem, Metrics, Shaping, SwashCache};
-use std::cmp::{self, Ordering};
-use termion::{color, cursor};
+use std::fmt::Write;
 
 fn main() {
     // A FontSystem provides access to detected system fonts, create one per application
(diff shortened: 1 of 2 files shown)


jackpot51:
This now has conflicts after #122

tigregalis:
I've resolved the conflict.

examples/terminal/src/main.rs
```diff
@@ -12,107 +15,67 @@ fn main() {
...
-    }
-    if (last_y as u16) < height {
-        print!("{}", cursor::Down(height - last_y as u16));
+    // Render the canvas
+    let mut output = String::new();
+
+    for row in canvas {
+        for pixel in row {
+            let (r, g, b) = pixel.unwrap_or((0, 0, 0));
+            write!(&mut output, "{}", "  ".on_truecolor(r, g, b)).ok();
```
nicoburns, line 75: This string needs to be one space instead of two, otherwise everything renders double sized

nicoburns (changes requested):
@jackpot51 Other than the comment I left above, I think this PR is good to go.

examples/terminal/src/main.rs
tigregalis: This was a deliberate choice as I wanted to see square-ish pixels, which is closer to a rendered image. Most monospace fonts people use in the terminal tend to be pretty narrow, approx 1:2.

Various terminal fonts using two spaces:
two-spaces

Various terminal fonts using one space:
one-space

jackpot51 (approved):

PR #136, opened 2023-06-08
allow setting cursor color

StaffEngineer:


jackpot51:
There is an issue with the build CI test, please take a look.

PR #138, opened 2023-06-08
allow setting cursor color

StaffEngineer:


jackpot51 (approved):

jackpot51:
This will be changed in #206, in order to make Eq work correctly for Cursor. Editor::draw will take the cursor color as an argument.

Issue #64, opened 2023-01-25
Spans of font sizes


rhelmot:
This is a highly requested feature from a rich text editor - being able to change font size within a line.

I looked into doing this myself but couldn't figure out a good way to get the attrs into layout without adding a font size field to the shaping structs, which seemed to be against the spirit of the design. If there's any advice on what changes would need to be made where to make this happen, I'll gladly do it.

jackpot51:
Color was also done in that sort of strange way. Perhaps both color and font size could be a struct specifically for overriding things later on in the process?

geom3trik:
How much later though? Because wouldn't font size affect shaping? (it my not, I'm just wondering)

jackpot51:
font size affects layout, not shaping. Though I'd probably break up runs of different font sizes into different shaping runs to make sure they don't have ligatures connecting them.

J-Cake:
I just wanted to leave a _plus one_ on this please. I would make heavy use of this

StaffEngineer:
Yeah, I am implementing markdown support for my app using cosmic-text, and for implementing markdown headings this feature would be really handy, so plus one to it.

StaffEngineer:
Hello, @conradludgate! I noticed that you referenced this specific issue and made some changes to the code. I'm curious to know the current status of those changes. Do you have any plans to submit a pull request (PR) soon to address the issue? I'm quite interested in having this feature implemented and would appreciate any updates or information you can provide. Thank you!

conradludgate:
> Hello, @conradludgate! I noticed that you referenced this specific issue and made some changes to the code. I'm curious to know the current status of those changes. Do you have any plans to submit a PR soon to address the issue? I'm quite interested in having this feature implemented and would appreciate any updates or information you can provide. Thank you!

I believe it worked well enough for my needs. I don't have time right now to continue working on it, but I'm happy for someone to take my commit and open a PR from it

StaffEngineer:
@conradludgate I gave the changes a try, but they're not quite what I had in mind. What I really need is a feature to make spans bigger by adjusting the font size, so I can effectively implement markdown headings.

PR #143, opened 2023-06-20
Offsetting and scaling

hecrj:
This PR introduces a new `PhysicalGlyph` type that allows deferring the computation of a `CacheKey` while offsetting and scaling a glyph after layout.

This allows consumers to benefit from layout linearity and leverage subpixel glyph positioning.

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -14,19 +14,24 @@ pub struct LayoutGlyph {
     pub start: usize,
     /// End index of cluster in original line
     pub end: usize,
+    /// Font size of the glyph
+    pub font_size: f32,
+    /// Font id of the glyph
+    pub font_id: fontdb::ID,
+    /// Font id of the glyph
+    pub glyph_id: u16,
     /// X offset of hitbox
     pub x: f32,
-    /// width of hitbox
+    /// Y offset of hitbox
+    pub y: f32,
+    /// Width of hitbox
     pub w: f32,
     /// Unicode BiDi embedding level, character is left-to-right if `level` is divisible by 2
     pub level: unicode_bidi::Level,
-    /// Cache key, see [CacheKey]
-    pub cache_key: CacheKey,
     /// X offset in line
     ///
-    /// If you are dealing with physical coordinates, you will want to use [`Self::x_int`]
-    /// together with [`CacheKey::x_bin`] instead. This will ensure the best alignment of the
-    /// rasterized glyphs with the pixel grid.
+    /// If you are dealing with physical coordinates, use [`Self::physical`] to obtain a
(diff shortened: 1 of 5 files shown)


src/layout.rs
```diff
@@ -14,44 +14,72 @@ pub struct LayoutGlyph {
...
+    /// Cache key, see [CacheKey]
+    pub cache_key: CacheKey,
+    /// Integer component of X offset in line
+    pub x: i32,
+    /// Integer component of Y offset in line
+    pub y: i32,
+}
+
+impl LayoutGlyph {
+    pub fn physical(&self, offset: (f32, f32), scale: f32) -> PhysicalGlyph {
```
hecrj, line 65: I don't really like the name. Ideally, it should be a verb; but I can't quite come up with anything. I'll keep thinking about it.

src/layout.rs
```diff
@@ -14,44 +14,72 @@ pub struct LayoutGlyph {
...
+        let x_offset = self.font_size * self.x_offset;
+        let y_offset = self.font_size * self.y_offset;
+
+        let (cache_key, x, y) = CacheKey::new(
+            self.font_id,
+            self.glyph_id,
+            self.font_size * scale,
+            (
+                (self.x + x_offset) * scale + offset.0,
+                ((self.y - y_offset) * scale + offset.1).trunc(), // Hinting in Y axis
```
hecrj, line 75: Maybe `hinting` could be an argument flag to this method and could be passed and stored in the `CacheKey`. This way we could tie it all together with the `swash::swash_image` function in a type safe way.

hecrj: From a Discord discussion, the author of `swash` shared:

> My general recommendation would be to never apply fractional offsets in the y direction. We do so in the x direction because scaled advance widths are almost never integral and we’re generally rasterizing to a fixed grid so if the fractional component of the advance is not taken into account, you end up with very noticeable inconsistent spacing between glyphs and the error accumulates quickly over a line.
> For most body text, the line height is consistent so rounding is fine (and imho, preferable).

So I guess we can just keep it hardcoded like this.

StaffEngineer:
Hello @hecrj! I noticed the presence of the font_size attribute on LayoutGlyph in the code. I'm curious to know if this addition will address the issue mentioned in https://github.com/pop-os/cosmic-text/issues/64 or not really?

hecrj:
@StaffEngineer No. `font_size` is on `LayoutGlyph` to defer the computation of the `CacheKey`.

jackpot51:
I fixed the test failures in https://github.com/pop-os/cosmic-text/commit/85ac47305664bc61273b578ec903636fca06ee94

PR #147, opened 2023-06-26
Adapt examples to be recognised as such by Cargo

WhyNotHugo:
Currently examples are binary modules inside the workspace. As a result, these are not treated as examples by cargo, e.g.:

```console
> cargo run --example
error: "--example" takes one argument.
No examples available.
```

This patch moves files around a bit so that examples are actually recognised as such by cargo. This has a few effects:

* `cargo test` will ensure that they compile.
* They will share the lock file with the main project. This ensure that their dependencies are aligned and avoids potentially downloading and compiling multiple different versions of the same dependency.
* `cargo run --example` provides useful output:

```console
> cargo run --example
error: "--example" takes one argument.
Available examples:
    editor-libcosmic
    editor-orbclient
    editor-test
    rich-text
    terminal
```

Depending on the shell's setup, shell completion may also be available when using `cargo run --example`.


WhyNotHugo:
CI seems to be failing because some of the dependencies used by examples have licenses that are not explicitly allowed: `BSD-3-Clause`, `Apache-2.0`, `BSL-1.0`, `BSD-2-Clause`, `ISC`, `MIT`, `OFL-1.1`.

I guess that the previous setup wasn't checking all licences correctly? Note that these are now development dependencies, _not_ dependencies that are built into the crate itself.

notgull:
-1 to this. Some of these examples have crates with deep dependency trees, and including those in every dev compilation would slow down my own development time.

WhyNotHugo:
These are only compiled when you run tests. I think that if you’re running tests you’d also want to make sure that examples haven’t broken.

Is the build time really even noticeable? Dependencies won’t be recompiled during rebuilds.

PR #159, opened 2023-07-22
Upgrade missing glyphs during shaping to a warning

notgull:
It is annoying to have to grep through DEBUG logs for this.


jackpot51:
This library should not be injecting log messages at this high level. I'd recommend making this a feature that js disabled by default

notgull:
> I'd recommend making this a feature that js disabled by default

Fixed!

jackpot51 (approved):

PR #160, opened 2023-07-24
Unify the no_std and std impls of FontSystem

notgull:
This would make it easier to add certain features (i.e. configurable font fallback) to both instances of the `FontSystem`.

This adds `hashbrown` and `rustc-hash` as dependencies for `no_std`.


notgull:
@jackpot51 If you have time can you review this change?

jackpot51 (approved):

grovesNL:
@notgull won't these dependencies always be enabled on std? I thought they'd be gated under `no_std`

notgull:
They do, but they're generally pretty lightweight dependencies.

grovesNL:
Yeah that's fair, I'm using std but intentionally try to avoid hashbrown when targeting wasm for size, so it's not ideal to include them by default.

notgull:
That's fair, but I'm not aware of a way to do this in stable Rust.

grovesNL:
What prevents us from gating the dependencies and hashmap alias to hashbrown under `no_std`? I'd be happy to submit a PR to adjust this but not sure if there's something I'm missing.

PR #161, opened 2023-07-25
Make it so fallback fonts can be added easily

### LINKED ISSUE #126: Allow setting the fallback fonts
I'm trying to add support for WASM to my `piet-cosmic-text` crate by embedding a font into the executable. However, it seems that, even after I set the font using `set_sans_serif_font` on the `fontdb::Database`, I get this panic:

```
panicked at 'no default font found', /home/jtnunley/.cargo/registry/src/github.com-1ecc6299db9ec823/cosmic-text-0.8.0/src/shape.rs:130:33
```

It would be nice to be able to set the fallback font to something such that it can be used in the event where none are available.

cc notgull/piet-cosmic-text#11

notgull:
Builds on #160, closes #126 by allowing the user to set their own fallback fonts.


notgull:
On second thought, this isn't the best way of doing it.

PR #162, opened 2023-07-25
Derive Clone for AttrsList

notgull:


notgull:
It turns out I don't need this for `piet-cosmic-text`... but it would still be nice.

jackpot51 (approved):

PR #170, opened 2023-08-15
Some code organization changes in `ShapeLine::layout`

Imberflur:
I was reading through this code (to get a better understanding of how alignment is applied and potential avenues for reusing layout work done for width measurement when performing the final layout) and noticed a few things that could be easily de-duplicated or adjusted to improve clarity (IMO).

* Some variables declared at the top of `ShapeLine::layout` weren't used until the second phase which is much farther down, so I moved them down to be at the start of that phase.
* Replaced the `push_line` `bool` with checking `layout_lines.is_empty()` which is effectively equivalent.
* Some variables defined outside the loop over `visual_lines` were reset with each loop iteration, so I moved their definitions down into the loop.
* Added a separate `justification_expansion` instead of reusing the `alignment_correction` variable (which just shifts the line in all other modes, but was used to hold how much each space should be expanded when justifying). This is more clear to the reader and avoids several later checks of whether the alignment is `Align::Justified` in spots where `alignment_correction` is used. _(I may have gone overboard with the other changes but I think at least this one is pretty useful)_
* I was able to factor out the common code during iterating visual line ranges for LTR/RTL text into a `process_range` closure as well as combine several cases of iterating over glyphs during this.
* Removed unnecessary `mem::swap(&mut glyphs, &mut glyphs_swap); ` (`glyphs` is not used later and can just be directly moved out of)

None of these changes should affect the behavior (If I didn't make any mistakes!).

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -875,17 +875,6 @@ impl ShapeLine {
         align: Option<Align>,
         layout_lines: &mut Vec<LayoutLine>,
     ) {
-        let align = align.unwrap_or({
-            if self.rtl {
-                Align::Right
-            } else {
-                Align::Left
-            }
-        });
-
-        // This is used to create a visual line for empty lines (e.g. lines with only a <CR>)
-        let mut push_line = true;
-
         // For each visual line a list of  (span index,  and range of words in that span)
         // Note that a BiDi visual line could have multiple spans or parts of them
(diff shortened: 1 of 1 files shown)


jackpot51:
This now conflicts after #175, please rebase.

jackpot51 (approved):

PR #174, opened 2023-08-23
enable fontconfig feature by default

### LINKED ISSUE #167: unable to find fonts on NixOS
# description
`FontSystem::new()` fails to find any fonts on NixOS.

# reproduction
```rust
fn main() {
  pretty_env_logger::init();
  cosmic_text::FontSystem::new();
}
```
output from `RUST_LOG=trace cargo run`:
```
 DEBUG cosmic_text::font::system::std > Locale: en-US
 INFO  cosmic_text::font::system::std > Parsed 0 font faces in 0ms
```

# suggested resolution
Enabling the `fontconfig` feature of `fontdb` seems to resolve the issue.

output with `fontconfig` feature enabled:
```
 DEBUG cosmic_text::font::system::std > Locale: en-US
 INFO  cosmic_text::font::system::std > Parsed 82 font faces in 18ms.
```

urlordjames:
Resolves #167.  See the discussion in RazrFalcon/fontdb#56 (TLDR: this isn't a perfect solution because fontdb's "`fontconfig` support isn't perfect... a proper Linux solution would require using `fontconfig` directly").

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ default-features = false
 features = ["hardcoded-data"]
 
 [features]
-default = ["std", "swash"]
+default = ["std", "swash", "fontconfig"]
 no_std = [
   "rustybuzz/libm",
   "hashbrown",


jackpot51 (approved):

tjw123hh:
https://github.com/RazrFalcon/fontdb/issues/56 suggests that we should entirely avoid using fontdb and directly use fontconfig to index fonts on Linux. Perhaps we should have a better solution?```

PR #178, opened 2023-09-10
Add IME support, extended actions, customize appearance

Riateche:
I was trying to implement a text input that works the same way as a native text input on desktop platforms. I would like to contribute the changes I had to make to `cosmic-text`.

# IME support

Input method editors are required for convenient text input in some languages, so its support is essential for good user experience. While it's possible to implement some of the following changes in a wrapper around `cosmic-text`, I think it would be better to incorporate them here and provide this support to all users.

1. Added `SetPreedit` action that inserts text with a special `is_preedit` flag in the attributes and clears any previous preedit. This is the most straightforward way to show a preedit inline, so that it works naturally with shaping and word wrapping. The downside is that the buffer's text is not the same as the document text anymore. For example, if the user wants to retrieve the text from the buffer, they should delete the part corresponding to the preedit (or clear the preedit). To help with this, I added `Buffer::text_without_preedit()` and `BufferLine::text_without_preedit()`.
2. Added `Edit::cursor_position()` method to retrieve the coordinates of the visual cursor. This is required to set the IME position, so that the input method window is displayed close to the cursor. The current implementation simply reuses the `Editor::draw()` code (moved out to a function). This can probably be done more efficiently.

There are also some aspects that I decided to implement outside of `cosmic-text` for now:

1. Handling mouse events while there is an active preedit. There are some applications where it's possible to set the IME cursor position by clicking inside the preedit area, but in the most cases the applications just ignore mouse clicks and drags initiated over the preedit area. However, if you click outside of the preedit, the current preedit is usually just committed, and the IME state is reset. This reset action is a problem because it means we cannot simply implement it in the editor's action handler.  There has to be a way for the editor to communicate to the host application that an IME reset is needed.  There are ways to modify the API for that, but I decided that it's out of scope of this PR.
2. Drawing underline of the preedit, which seems to be the most common way to represent it visually. It's possible to add support for underline styling to `Buffer` (for example, by integrating with `line-straddler`), at which point it would be pretty easy to change the default preedit attributes to underline instead of the current color change.

# Hiding text cursor

1. Cursor is now hidden when there is a selection.
2. Cursor is now hidden when there is an IME preedit that specifies that there should be no cursor.
3. Added `Edit::set_cursor_hidden()` to hide cursor manually, which can be used to implement cursor blinking, and to hide cursor when the input is unfocused or read only.

While I generally prefer "positive" boolean flags (e.g. `set_cursor_visible()` would generally be better), in this case the logic looks simpler when we use the `hidden` flag: `cursor_hidden = has_selection || has_preedit_without_cursor || set_cursor_hidden`.

# Extended mouse and keyboard actions

I was aiming to implement all standard keyboard shortcuts and mouse actions that are usually present on desktop systems.

1. Added missing actions: `DocumentStart`, `DocumentEnd`, `SelectAll`, `DeleteStartOfWord`, `DeleteEndOfWord`, `SelectWord`, `SelectParagraph`.
2. Added `select` flag to `Edit::action()` to implement hotkeys with `Shift` modifier (e.g. `Shift + ArrowRight` to create or extend the selection) and fix hotkeys without `Shift` (e.g. `ArrowRight` previously didn't clear existing selection). I think this is the simplest way to implement it currently. It's not ideal because this flag only makes sense for some actions, but refactoring this API would be more dirsuptive.
3. Fixed `Previous` and `Next` behavior when there is a selection. They now clear selection and put the cursor to the start and end of the selection, respectively.
4. Fixed `NextWord` action to go to the end of line if there are no more word boundaries. Previously, it could get stuck and not move forward at all in some cases.

All new actions can be tested in the `editor-orbclient` example (I chose it because it already had some keyboard modifiers support, unlike the other examples).

# More styling options

1. Added an option to change the selection color (it was previously hardcoded).
2. Added an option to set the selected text color. This is often needed to replicate the appearance of native widgets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ modit = { version = "0.1.4", optional = true }
 rangemap = "1.4.0"
 rustc-hash = { version = "1.1.0", default-features = false }
 rustybuzz = { version = "0.12.0", default-features = false, features = ["libm"] }
+itertools = "0.11.0"
 self_cell = "1.0.1"
 swash = { version = "0.1.12", optional = true }
 syntect = { version = "5.1.0", optional = true }
--- a/examples/editor-test/src/main.rs
+++ b/examples/editor-test/src/main.rs
@@ -17,6 +17,7 @@ fn redraw(
     let font_color = Color::rgb(0xFF, 0xFF, 0xFF);
     let cursor_color = Color::rgb(0xFF, 0xFF, 0xFF);
     let selection_color = Color::rgba(0xFF, 0xFF, 0xFF, 0x33);
+    let selected_text_color = Color::rgb(0xF0, 0xF0, 0xFF);
 
     editor.shape_as_needed(true);
     if editor.redraw() {
@@ -29,6 +30,7 @@ fn redraw(
             font_color,
             cursor_color,
             selection_color,
+            selected_text_color,
             |x, y, w, h, color| {
                 window.rect(x, y, w, h, orbclient::Color { data: color.0 });
             },
--- a/examples/editor-test/src/main.rs
+++ b/examples/editor-test/src/main.rs
@@ -116,7 +116,10 @@ fn main() {
             // Test delete of EGC
             {
                 let cursor = editor.cursor();
-                editor.action(Action::Motion(Motion::Previous));
+                editor.action(Action::Motion {
+                    motion: Motion::Previous,
+                    select: false,
+                });
                 editor.action(Action::Delete);
                 for c in grapheme.chars() {
                     editor.action(Action::Insert(c));
@@ -140,7 +143,10 @@ fn main() {
         {
             let cursor = editor.cursor();
             editor.action(Action::Enter);
-            editor.action(Action::Motion(Motion::Previous));
+            editor.action(Action::Motion {
+                motion: Motion::Previous,
+                select: false,
+            });
             editor.action(Action::Delete);
             assert_eq!(cursor, editor.cursor());
         }
--- a/examples/editor/src/main.rs
+++ b/examples/editor/src/main.rs
@@ -63,6 +63,7 @@ fn main() {
     }
 
     let mut ctrl_pressed = false;
+    let mut shift_pressed = false;
     let mut mouse_x = 0.0;
     let mut mouse_y = 0.0;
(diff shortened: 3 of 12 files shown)


Riateche:
I would appreciate any feedback on this.

jackpot51:
This needs to be rebased to resolve conflicts first

Riateche:
I rebased and updated the PR (it wasn't easy after half a year :D). Thanks to the added separation between `Action` and `Motion`, the `select` flag implementation is now much cleaner. I also updated the `editor` example to showcase all the new features. Please take a look.

Riateche:
I fixed the `cargo fmt` check, but not sure what I can do about the double syn dependency failure in CI. It seems unrelated to the PR.

Riateche:
Let me know if I can do something to help move this forward.

jackpot51:
This is a large set of changes, so it is admittedly hard for me to review and approve. If it could be split into a few more PRs, that would be helpful. There are changes that are more likely to be merged as-is and other changes that probably need some refactoring.

Riateche:
Sure! Here are the separated PRs:
- https://github.com/pop-os/cosmic-text/pull/261
- https://github.com/pop-os/cosmic-text/pull/262
- https://github.com/pop-os/cosmic-text/pull/263
- https://github.com/pop-os/cosmic-text/pull/264
- https://github.com/pop-os/cosmic-text/pull/265

Once we're done with these ones, I'll open PRs with the remaining changes.

jackpot51:
> Sure! Here are the separated PRs:
> 
>     * [Add IME preedit ranges #261](https://github.com/pop-os/cosmic-text/pull/261)
> 
>     * [Add Edit::cursor_position #262](https://github.com/pop-os/cosmic-text/pull/262)
> 
>     * [Hide cursor based on selection, preedit, or config #263](https://github.com/pop-os/cosmic-text/pull/263)
> 
>     * [Add an option to set selected text color #264](https://github.com/pop-os/cosmic-text/pull/264)
> 
>     * [Add select flag for motion and click actions #265](https://github.com/pop-os/cosmic-text/pull/265)
> 
> 
> Once we're done with these ones, I'll open PRs with the remaining changes.

Thanks! I will review when I am back at work on Monday.

PR #183, opened 2023-09-17
Set MSRV to 1.65

EHfive:
The MSRV was scanned by cargo-msrv, the result shows src/buffer.rs uses let - else statements which requires rust 1.65.0. This also reformat the Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 license = "MIT OR Apache-2.0"
 documentation = "https://docs.rs/cosmic-text/latest/cosmic_text/"
 repository = "https://github.com/pop-os/cosmic-text"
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

alerque:
Just a heads up, MSRV bumps mandate a *major* version bump of the crate. I'm not sure how to handle the case where the defacto minimum in a previous release was already higher and the value just wasn't specified properly. That might not be considered a breaking change because anybody that could compile it before still can, but in the event the last release didn't have the let/else statements at all and the new MSRV is actually higher than the last release, the major version should be incremented. Since you are in zero-dot territory I guess you are making up your own versioning rules to some extent, but in a 1.x.y or greater library an MSRV bump would necessitate a 2.x.y release.

PR #184, opened 2023-09-17
Define a macro to create self-referencing type and use it for font face

EHfive:
This makes unsafe part be more concentrated and allows reuse. The macro would be used in later patch.

### DIFF
--- a/src/font/mod.rs
+++ b/src/font/mod.rs
@@ -1,109 +1,54 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pub(crate) mod fallback;
 
-use alloc::boxed::Box;
+use core::fmt;
+
 use alloc::sync::Arc;
 
 pub use self::system::*;
 mod system;
 
-pub use font_inner::Font;
+mod owned_face {
+    impl_self_ref!(OwnedFace, rustybuzz::Face<'static>, rustybuzz::Face<'this>);
+}
+use owned_face::*;
 
-/// Encapsulates the self-referencing `Font` struct to ensure all field accesses have to go through
-/// safe methods.
-mod font_inner {
-    use super::*;
-    use aliasable::boxed::AliasableBox;
-    use core::fmt;
+pub struct Font {
+    #[cfg(feature = "swash")]
+    swash: (u32, swash::CacheKey),
+    rustybuzz: OwnedFace<Arc<dyn AsRef<[u8]> + Send + Sync>>,
+    data: Arc<dyn AsRef<[u8]> + Send + Sync>,
+    id: fontdb::ID,
+}
 
-    /// A font
-    //
-    // # Safety invariant
-    //
-    // `data` must never have a mutable reference taken, nor be modified during the lifetime of
-    // this `Font`.
-    pub struct Font {
-        #[cfg(feature = "swash")]
-        swash: (u32, swash::CacheKey),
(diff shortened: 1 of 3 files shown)


hecrj:
Have we considered using `ouroboros` instead of rolling our own unsafe?

EHfive:
> Have we considered using `ouroboros` instead of rolling our own unsafe?

There was, but dropped for compilation time reason. https://github.com/pop-os/cosmic-text/commit/95e36249d51bb1d31e84e38bbfe46a95e8f33006 Though it's not a big issue for me..

The produced type is basically equivalent to expansion of ouroboros macro so I think it's fine. 

And this macro should still be faster than ouroboros as we didn't use proc-macro here.

Imberflur:
Imo if this will be using custom unsafe, then it is easier to verify the soundness of a particular implementation rather than a generic macro. And I don't see any other motivation for a macro in this case other than encapsulating the unsafe code.

Also, the current implementation has a soundness issue, but it is not too surprising since this was just fixed in ouroboros <https://github.com/joshua-maros/ouroboros/issues/88>. I can make a PR for this later today.

src/font/mod.rs
```diff
@@ -1,109 +1,54 @@
...
 use alloc::sync::Arc;
 
 pub use self::system::*;
 mod system;
 
-pub use font_inner::Font;
+mod owned_face {
+    impl_self_ref!(OwnedFace, rustybuzz::Face<'static>, rustybuzz::Face<'this>);
+}
+use owned_face::*;
```
Imberflur, line 14: wrapping this up in a module is necessary to prevent accessing the fields incorrectly so it should probably be part of the macro

EHfive:
@Imberflur After roughly look at the issue you mentioned, and given the motivation of not using `ouroboros` is to speed up compilation. I think we should just use the [self_cell](https://github.com/Voultapher/self_cell) which is proc-macro-free thus being faster to compile. And to avoid implementing indigenous unsafe code here which has less change to be audited.

src/macros.rs
```diff
@@ -0,0 +1,69 @@
...
+/// ```
+macro_rules! impl_self_ref {
+    ($SelfRef:ident, $RefStatic:ty, $RefThis:ty) => {
+        /// # Safety invariant
+        ///
+        /// `data` must never have a mutable reference taken, nor be modified during the lifetime of
+        /// this struct
+        pub struct $SelfRef<T> {
+            /// `data_ref` could self-referencing `data`
+            data_ref: $RefStatic,
```
Imberflur, line 26: This field needs to be wrapped in `maybe_dangling::MaybeDangling` or `core::mem::MaybeUninit` to prevent the compiler from potentially assuming any references it contains will be valid till the end of a function when `$SelfRef<T>` is passed as a function parameter (see the issue I linked above). Any such references will only be valid until this value is dropped.

Imberflur:
> @Imberflur After roughly look at the issue you mentioned, and given the motivation of not using `ouroboros` is to speed up compilation. I think we should just use the [self_cell](https://github.com/Voultapher/self_cell) which is proc-macro-free thus being faster to compile. And to avoid implementing indigenous unsafe code here which has a lesser chance to be audited.

That definitely seems like the best approach. We could have removed the extra layer of indirection from the `AliasableBox`, since an `Arc` already wraps the inner data but I think it is much more valuable to rely on unsafe code the will have a higher chance of being well audited.

PR #185, opened 2023-09-17
Implement no_std compatible ab_glyph glyph rendering backend

EHfive:
> [!NOTE] 
> ~~This PR depends on #183 and #184 . Please review & merge those first.~~ Merged or have alternative #188 .

### Implement no_std compatible ab_glyph glyph rendering backend

This uses ab_glyph to scale & rasterize glyph outline and load glyph image, it
can be served as an alternative to swash.

For glyph image(e.g. color emoji) rendering, only support for images in PNG
format are currently implemented due to upstream limitation.

<details>
<summary>Screenshot of demo targeting no_std x86_64-unknown-uefi</summary>

Qemu UEFI

</details>

---

### Add ab_glyph feature to examples and use the new draw_with method

Run examples with `cargo run --features ab_glyph` to test ab_glyph rendering.

PR #189, opened 2023-09-19
Abstract glyph drawing logic as Draw trait

EHfive:
> [!NOTE] 
> This is separated from #185 for easier review.

The trait contains only one method, draw_line, which is going to be implemented by glyph rendering backend like swash to scale & rasterize inputed glyphs and draw pixels of them.

This make it easier to implement other rendering backend, and allows switching rendering backend at runtime.

The original code of glyph rendering for swash is moved into Draw.

And add draw_with method to Buffer and Edit interfaces to draw with glyph rendering backend that implemented Draw trait. The original glyph drawing part using swash in buffer and editor is also migrated to using Draw.

---

I am thinking we could also deprecate swash specific `draw` methods in Buffer and Edit to use `draw_with` instead. Or just modify the interface of `draw`.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -9,7 +9,7 @@ use core::{cmp, fmt};
 use unicode_segmentation::UnicodeSegmentation;
 
 use crate::{
-    Attrs, AttrsList, BidiParagraphs, BorrowedWithFontSystem, BufferLine, Color, FontSystem,
+    Attrs, AttrsList, BidiParagraphs, BorrowedWithFontSystem, BufferLine, Color, Draw, FontSystem,
     LayoutGlyph, LayoutLine, ShapeBuffer, ShapeLine, Shaping, Wrap,
 };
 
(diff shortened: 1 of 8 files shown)


notgull:
I think it would be better to just implement `no_std` support for Swash than to add the complexity of pluggable drawing backends

EHfive:
@notgull
Some benefits for abstracting Draw trait.  
- There was 2 copies of swash rendering code in Buffer and Editor, this helps to make the code be more organized.
- From `SwashCache` user's perspective, the `draw` and `draw_with` is the same for calling except the method name.
- If we decide to make `Draw` public(as it currently is), external crates are allowed to extend Buffer and Edit with their own glyph rendering backend.

Regarding `no_std`, I am currently in needs of `no_std` capability for my own [hobby project](https://github.com/EHfive/uefi-toys) and I found it easier for me to work on crates that are already `no_std`.

Issue #190, opened 2023-09-23
Hebrew words (RTL) are not rendered correctly on main


dovreshef:
Hi,

It appears to be a regression introduced by commit (ae96bf26d1a5d9a309139256ed9e5eb6606f7232).

This can be demonstrated in the following test case:
```rust
use cosmic_text::{
    Attrs,
    Buffer,
    Color,
    FontSystem,
    Metrics,
    Shaping,
    SwashCache, Family,
};
use tiny_skia::{
    Paint,
    Pixmap,
    Rect,
    Transform,
};

fn main() {
    // A FontSystem provides access to detected system fonts, create one per application
    let mut font_system = FontSystem::new();

    // A SwashCache stores rasterized glyphs, create one per application
    let mut swash_cache = SwashCache::new();

    // Text metrics indicate the font size and line height of a buffer
    let metrics = Metrics::new(36.0, 50.0);

    // A Buffer provides shaping and layout for a UTF-8 string, create one per text widget
    let mut buffer = Buffer::new(&mut font_system, metrics);

    // Borrow buffer together with the font system for more convenient method calls
    let mut buffer = buffer.borrow_with(&mut font_system);

    // Set a size for the text buffer, in pixels
    buffer.set_size(400.0, 100.0);

    // Attributes indicate what font to choose
    let attrs = Attrs::new();

    // Add some text!
    buffer.set_text("בדיקות", attrs, Shaping::Advanced);

    // Perform shaping as desired
    buffer.shape_until_scroll();

    // Create a default text color
    let text_color = Color::rgb(0x00, 0x00, 0x00);

    let mut pixmap = Pixmap::new(500, 500).unwrap();
    pixmap.fill(tiny_skia::Color::WHITE);

    // Draw the buffer (for performance, instead use SwashCache directly)
    buffer.draw(&mut swash_cache, text_color, |x, y, w, h, color| {
        let mut paint = Paint {
            anti_alias: false,
            ..Default::default()
        };
        paint.set_color_rgba8(color.r(), color.g(), color.b(), color.a());
        let rect = Rect::from_xywh(x as f32, y as f32, w as f32, h as f32).unwrap();
        pixmap.fill_rect(rect, &paint, Transform::identity(), None);
    });

    pixmap.save_png("test.png").unwrap();
}
```

The test uses tiny-skia to render a hebrew word to a PNG file. Use the dependencies below to replicate:
```
[dependencies]
[...]

PR #191, opened 2023-09-23
Fix rendering of RTL words

dovreshef:
Fixes (I think, my understanding of the code is very limited) issue #190.

A related question:

I see that there are no tests for these type of issues. 

I can add a few tests for Hebrew, and English. My thinking is to use an open source font to render an image of a word or a paragraph to an image, and compare it against a master image saved in the repository. That should catch these kinds of issues at least..

Would that be desired/helpful?

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -1221,11 +1221,12 @@ impl ShapeLine {
                                 } else {
                                     0.0
                                 };
-                            let y_advance = font_size * glyph.y_advance;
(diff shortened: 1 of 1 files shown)


jackpot51:
Thanks for the PR! The test you describe would be perfect, using an open source font and requiring stable output as part of cargo test.

jackpot51 (approved):

PR #192, opened 2023-09-23
Add image render tests

dovreshef:
continuing from #191.

Add tests that will match rendered words/paragraphs against reference images.

Use env var `GENERATE_IMAGES` to write the initial reference images to the repository.

NOTE: The images I'm rendering looks blurry for some reason.. I'm not sure what exactly I'm doing wrong.


CryZe:
An alternative to storing the reference images could be to just hash the result and see if the hash still matches.

dovreshef:
> An alternative to storing the reference images could be to just hash the result and see if the hash still matches.

@CryZe That's in some way much better, since then there is no need for large images and comparing them.
Although, it might make it harder to debug what's the issue if the test fails.

Anyway will leave it to you? @jackpot51? decision.

Do you also maybe know why the text is coming out blurry in the image?

jackpot51:
You could use Noto Sans Hebrew and/or Noto Sans Arabic for RTL testing, these are licensed under the SIL Open Font License. I prefer images, since it makes it possible to compare the results and see exactly what the failure is. They can be stored with GIT LFS, along with the font files, and in their current PNG compression take up very little space.

dovreshef:
@jackpot51 Moved to using Noto Sans Hebrew + Arabic and added Arabic tests.

Three notes:

1. I used Google translate for the Arabic since I don't speak it. so I translated "test" and "the quick brown fox jumps over the lazy dog" into Arabic for the tests. I used those phrases for the Hebrew as well.
2. I could not add a test for mixed Arabic/English since Noto Sans Arabic does not include Latin fonts? (got bounding boxes for the English stuff).
3. The Hebrew rendering of Noto Sans Hebrew does not look as good as the other font. But I'm not sure if it's a bug in rendering, the font, or something else. Specifically, the issue is that some letters are too close together and other a bit too far apart.

Additionally cargo-deny is failing because tiny-skia (which I've added as a dev dependency) is using the "BSD-3-Clause" license, which is not approved.

jackpot51:
> @jackpot51 Moved to using Noto Sans Hebrew + Arabic and added Arabic tests.
> 
> Three notes:
> 
>     1. I used Google translate for the Arabic since I don't speak it. so I translated "test" and "the quick brown fox jumps over the lazy dog" into Arabic for the tests. I used those phrases for the Hebrew as well.

It looks fine to me, thanks!


>     2. I could not add a test for mixed Arabic/English since Noto Sans Arabic does not include Latin fonts? (got bounding boxes for the English stuff).

I would recommend to add the Noto Sans font for Latin, Cyrillic, and Greek characters to the font database: https://fonts.google.com/noto/specimen/Noto+Sans.

>     3. The Hebrew rendering of Noto Sans Hebrew does not look as good as the other font. But I'm not sure if it's a bug in rendering, the font, or something else. Specifically, the issue is that some letters are too close together and other a bit too far apart.

Can you replicate this rendering with some text editor using Noto Sans Hebrew?

fonts/.DS_Store
jackpot51, line 1: It would be good to add `.DS_Store` to the `.gitignore`

dovreshef: Done.

dovreshef:
Added the font for Noto Sans English and started loading all fonts in the directory. Now the test for mixed English & Arabic works.

Regarding the font rendering, I'll paste a comparison below
image

I think it's just the way I'm using tiny-skia which is sub-par and blurry.

fonts/NotoSans-Regular.ttf
jackpot51, line 1: Make sure to add .ttf files to git lfs with:

`git lfs track '*.ttf'`

dovreshef: Done.

jackpot51 (approved):
This looks perfect to me now. I appreciate the help, this will go a long way to ensuring there are no future regressions!

PR #200, opened 2023-10-31
Bump `fontdb` to `0.16`

MarijnS95:
We were getting duplicate `ttf-parser` dependencies because of `rustybuzz` releasing its `ttf-parser 0.20` upgrade out of sync with `fontdb 0.16`, whose update just got published.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/pop-os/cosmic-text"
 rust-version = "1.65"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

MarijnS95:
Looks like some new crate releases fixed the `indexmap`/`hashbrown` duplicate.

Note though that this only checks the core `cosmic-text` crate (probably intended), but I get a whole bunch of errors when running `cargo deny --workspace check`.

jackpot51 (approved):

PR #206, opened 2023-12-15
Refactor

jackpot51:
Includes various performance improvements, refactor of cursor motion so it can be done outside of Editor, refactoring of scroll so that previous lines do not have to be shaped, and the ability to prune caches to reduce memory usage.

The most important change is to allow shaping parts of a buffer, by changing scroll to include both an index in Buffer::lines and the layout line. This lets only the part currently within the scrolled view of the buffer to be shaped and laid out. This greatly improves performance when scrolling as it does not have to shape and lay out lines that were never shown.

The shaping functions for Buffer and Editor now take a `prune` parameter, which when set to true, will drop the shaping and layout information of all lines outside of the scrolled view. This greatly decreases memory usage at the cost of reshaping when scrolling.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ self_cell = "1.0.1"
 swash = { version = "0.1.8", optional = true }
 syntect = { version = "5.1.0", optional = true }
 sys-locale = { version = "0.3.1", optional = true }
+ttf-parser = "0.20.0"
 unicode-linebreak = "0.1.5"
 unicode-script = "0.5.5"
 unicode-segmentation = "1.10.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/pop-os/cosmic-text"
 rust-version = "1.65"
 
 [dependencies]
+bitflags = "2.4.1"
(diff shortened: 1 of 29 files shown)


nicoburns:
Ah, it's just clicked for me that "scrolling" in cosmic text is supposed to be understood in the context of something like an editor where you have a fixed-sized area (not dependent on the content) and you can approximate scroll height without shaping by counting newlines. It previously hadn't made any sense to me why you'd only want to shape until scroll, or why cosmic text was involved with scrolling at all.

Somewhat relatedly to this, I've been thinking about what it would take to implement support for CSS floats on top of cosmic-text (on the assumption that cosmic-text would be unlikely to want to support them itself). And my conclusion so far is that it would require the following:
1. The ability to layout around excluded areas (rectangles)
2. An incremental/resumable text layout API where you can:
    - Layout up to a point (character index)
    - Yield to the caller and allow them to mutate both the layout generated so far and the input parameters (with them being required to only modify inputs that haven't been consumed so far if they want a sensible resulting layout)
    - Resume layout passing the mutated layout and inputs
    
This would allow yielding at the locations of floated content, running some external logic to insert a new excluded area and move placed text runs so that they do not overlap this area, and then resuming text layout from that point.

Issue #177, opened 2023-09-07
Identify projects using cosmic-text


jackpot51:
It would be nice to be able to look at other projects using cosmic-text.

jackpot51:
[lapce](https://github.com/lapce/lapce) appears to use a forked [cosmic-text](https://github.com/lapce/cosmic-text) for the [floem UI library](https://github.com/lapce/floem)

jackpot51:
[iced](https://github.com/iced-rs/iced) uses version 0.9 of cosmic-text

jackpot51:
[bevy](https://github.com/bevyengine/bevy) has [a PR open](https://github.com/bevyengine/bevy/pull/8808) for cosmic-text integration, using version 0.8

notgull:
[`theo`](https://github.com/notgull/theo) uses `cosmic-text` for text rendering.

mmstick:
[bevy-cosmic-edit](https://github.com/StaffEngineer/bevy_cosmic_edit) exists for unofficial bevy support.

CryZe:
[livesplit-core](https://github.com/LiveSplit/livesplit-core) as the default text engine of its renderer.

nicoburns:
[vizia](https://github.com/vizia/vizia) is using cosmic text for it's text support.

mmstick:
[glyphon](https://github.com/grovesNL/glyphon) uses version 0.9

mmstick:
[rusty-spine](https://github.com/jabuwu/rusty_spine) uses version 0.9

nicoburns:
[inlyne](https://github.com/trimental/inlyne) markdown preview app is using a fork of cosmic text

ecton:
[cushy](https://github.com/khonsulabs/cushy/) and it's wgpu-based rendering library [Kludgine](https://github.com/khonsulabs/kludgine) are both using v0.10 (and hope to never switch text rendering ibraries again! :laughing:)

noverby:
The Zed developers are considering using cosmic-text: https://github.com/zed-industries/zed/issues/7015

noverby:
Zed merged a PR with cosmic-text support: https://github.com/zed-industries/zed/pull/7539

msparkles:
yakui will be using cosmic text once https://github.com/SecondHalfGames/yakui/pull/179 gets merged (one day :tm: ) :)

AustinMReppert:
https://github.com/oku-gui/oku-gui

[...]

PR #207, opened 2023-12-19
Cache `rustybuzz` shape plans for vastly improved performance.

vorporeal:
I added support to `rustybuzz` to cache shape plans (https://github.com/RazrFalcon/rustybuzz/pull/88), which has been released in `rustybuzz` v0.12.0.

This PR updates `cosmic-text` to make use of the new `rustybuzz` functionality by adding a `ShapePlanCache`, storing shape plans (with an appropriate cache key), and passing the cached shape plan to `rustybuzz::shape_with_plan` (instead of `rustybuzz::shape`) to avoid re-paying the cost of computing the shape plan.

A couple notes:
* I renamed `cache.rs` to `glyph_cache.rs` now that there's a `shape_plan_cache.rs`, but kept the names of symbols within it generic (e.g.: `Cache`) as they're public outside of the crate.
* The shape plan cache does not have a bounded size.  I believe this is ok, due to the fact that:
  1. Nothing sets a language on the shaping buffer,
  2. I would expect there to only be one direction per script value (not to mention the fact that the max multiplication factor is 2), and
  3. I would expect there to usually be only one script value per font (e.g.: Latin/Hebrew/Chinese).
* Exposing the `ShapePlanCache` via `FontFallbackIter` is nasty, but there isn't a great alternative given the fact that `ShapePlanCache` is held by `FontSystem` and `FontFallbackIter` holds a `&mut FontSystem` and we need a `&mut ShapePlanCache` in `shape_fallback`.  That said, this nastiness is narrowly scoped and entirely internal to this crate.

All tests in `cosmic-text` still pass with this change.

For some anecdotal evidence on performance improvements - all font shaping/layout performance issues we were seeing in our app (https://github.com/warpdotdev/Warp) have disappeared with this patch.


jackpot51:
Thanks!

jackpot51:
If cosmic-text is being used by warp please consider mentioning this in #177

PR #211, opened 2024-01-15
Avoid wrapping a space onto its own newline

### LINKED ISSUE #155: `Wrap::Word` inserts newline for trailing space (after word)
If we have text like `Testing Spacing` in the rich-text example, you can see that the space causes a full empty line to be inserted (tested with the rich-text example):

image

(the cursor also does an interesting thing in this situation)

The behavior seems to have changed in #78. Before #78 this would look like:

image

The previous behavior seems better for display, because I wouldn't expect a space to wrap a whole line. Instead I'd expect the space to kind of bind to the word before it, so that "Testing " remains on the first line.

If the old behavior is preferred, I'll try to submit a PR to change how we wrap in this case.

The current handling for word wrapping is located in https://github.com/pop-os/cosmic-text/blob/aa1b37aede5435067bc51667797f16ffb8b8be7d/src/shape.rs#L831-L873 and https://github.com/pop-os/cosmic-text/blob/aa1b37aede5435067bc51667797f16ffb8b8be7d/src/shape.rs#L923-L968

grovesNL:
Fixes #155

There was some behavior added in #78 that intentionally seems to break trailing spaces in this way, but it doesn't seem be necessary in the test cases I've tried (mostly just experimenting with the rich-text example).

If anyone has a test case that this regresses, please let me know! As far as I can tell this is an improvement for typical cases.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -944,7 +944,6 @@ impl ShapeLine {
         } else {
             for (span_index, span) in self.spans.iter().enumerate() {
                 let mut word_range_width = 0.;
-                let mut width_before_last_blank = 0.;
                 let mut number_of_blanks: u32 = 0;
 
                 // Create the word ranges that fits in a visual line
@@ -967,7 +966,6 @@ impl ShapeLine {
                             // fits
                             if word.blank {
                                 number_of_blanks += 1;
-                                width_before_last_blank = word_range_width;
                             }
                             word_range_width += word_width;
                             continue;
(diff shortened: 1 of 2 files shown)


grovesNL:
`cargo-deny` failure is unrelated - caused by multiple version of `syn` and `bitflags`

jackpot51:
@grovesNL unfortunately that does look like a regression. Hopefully it is not too hard to fix.

alokedesai:
@grovesNL lmk if there's anyway I can help get this PR past the finish line. 

I just spent quite a few quite a few hours debugging a text layout issue in [Warp](https://github.com/warpdotdev/warp) that ultimately seems to be caused by this. Would love to see this fixed and happy to lend a hand if it would useful.

grovesNL:
@alokedesai absolutely, any help looking into this would be great! I was planning to step through the positioning in a debugger to see how we end up with two spaces there but might not have time until another week or so.

Based on the test case failure it seems like there's some kind of unexpected interaction when we have left-to-right and right-to-left text spans that end up on the same visual line.

PR #212, opened 2024-01-17
RFC: Try harder to succeed at fall-backing to a Monospace font

MoSal:
~~Depends on RazrFalcon/fontdb#62~~

A combination of ideas:

 * Try all Monospace fonts before giving up.
 * Relax exact weight restriction on font matching when trying Monospace fall-back. Try smaller weights if needed.
 * Make the fall-back try order weight-offset aware, AND script-aware.
 * And finally, add the option to adjust the font size of glyphs using fall-back Monospace fonts, so the width of them matches the default font width.

   For my use-case, the current fall-back attempt always fails with Arabic script. And none of the Arabic-supporting Monospace fonts in my system support medium weight. So, if my default font is set to medium weight, script-aware fall-back alone will still not work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ self_cell = "1.0.1"
 swash = { version = "0.1.8", optional = true }
 syntect = { version = "5.1.0", optional = true }
 sys-locale = { version = "0.3.1", optional = true }
+ttf-parser = "0.20.0"
 unicode-linebreak = "0.1.5"
 unicode-script = "0.5.5"
 unicode-segmentation = "1.10.1"
--- a/src/attrs.rs
+++ b/src/attrs.rs
@@ -177,7 +177,8 @@ impl<'a> Attrs<'a> {
         //TODO: smarter way of including emoji
         face.post_script_name.contains("Emoji")
             || (face.style == self.style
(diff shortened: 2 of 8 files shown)


MoSal:
`cosmic-term` showcases (default font set to Fira Mono + Medium weight):

# Non-Monospace fall-back

ct_no_mono

Notice how the width is different in the seconds line.

Wrapping is also broken:

ct_no_mono_broken_wrap

# Monspace fall-back (no resize)

Three different fonts:

## Dejavu

ct_no_resize_dejavu

## Cascadia

ct_no_resize_cascadia

## Kawkab Mono

ct_no_resize_kawkab

Dejavu only looked right because it happened to have the same width as Fira Mono. The broken wrapping (should be fixed in `cosmic-term` irregardless of this RFC) is particularly bad with Kawkab Mono because it's width is higher than Fira Mono (0.7 vs 0.6).

# Monospace fall-back (with resize)

## Dejavu

ct_fixed_dejavu


## Cascadia

ct_fixed_cascadia

## Kawkab Mono

ct_fixed_kawkab

Looks perfect.

MoSal:
No longer needs fontdb changes.

jackpot51 (approved):

PR #217, opened 2024-01-22
Add a size limit to `font_matches_cache` and empty it when it's reached

MoSal:
In `FontSystem`, `font_matches_cache` is an ever growing cache.

 It can also be a fast growing one in stress tests like running this
 in `cosmic-term`:

     mpv -speed 3 -vo tct <some_video>

 So this commit adds a size limit to that cache, and simply empties
 it when that limit is reached, which shouldn't be a common occurrence
 in normal usage.

### DIFF
--- a/src/font/system.rs
+++ b/src/font/system.rs
@@ -1,4 +1,4 @@
-use crate::{Attrs, AttrsOwned, Font, HashMap, ShapePlanCache};
+use crate::{Attrs, Font, FontMatchAttrs, HashMap, ShapePlanCache};
 use alloc::string::String;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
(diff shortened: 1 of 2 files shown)


MoSal:
There is another source of *leakage* with that stress test when a software renderer is involved. That one should be fixed by iced-rs/iced#2210.

src/font/system.rs
```diff
@@ -132,6 +133,12 @@ impl FontSystem {
     }
 
     pub fn get_font_matches(&mut self, attrs: Attrs<'_>) -> Arc<Vec<FontMatchKey>> {
+        // Empty the cache first if it reached the size limit
+        if self.font_matches_cache.len() >= Self::FONT_MATCHES_CACHE_SIZE_LIMIT {
+            log::trace!("empty font mache cache");
+            self.font_matches_cache = Default::default();
```
jackpot51, line 139: I think it is better to use .clear so it can keep capacity.

src/font/system.rs
```diff
@@ -132,6 +133,12 @@ impl FontSystem {
...
+        }
+
         self.font_matches_cache
             //TODO: do not create AttrsOwned unless entry does not already exist
             .entry(AttrsOwned::new(attrs))
```
jackpot51, line 144: I just realized this will insert a new item if something unrelated to finding fonts changes (like the metadata or color)

MoSal: Indeed.

Do you think it's worth it to change the map key to only contain relevant `Attrs` info?

jackpot51: I do think that is important. Can make a new struct specifically for this map's key

MoSal: Done.

jackpot51 (approved):

PR #224, opened 2024-01-31
Use weight absolute difference in monospace fallback matching

MoSal:
When matching on weights smaller than normal, "equal or smaller"
 weight restriction may cause monospace fallback to fail, depending
 on font support at such weights for the text to be shaped.

 So remove that restriction, and calculate weight differences instead
 of offsets.

 In case of no exact weight match, and with all other factors being
 equal, smaller weights will be picked before bigger ones. So, this
 should generally not cause any behavioral changes when matching on
 normal weight or bigger.

 Should fix pop-os/cosmic-term#104.


snaggen:
I can confirm that this do fix https://github.com/pop-os/cosmic-term/issues/104

jackpot51 (approved):

PR #225, opened 2024-02-02
Word wrap glyph fallback

dtzxporter:
This is an attempt to implement glyph wrapping fallback when a word can't be placed on a line by itself. It follows how the web handles it which starts the word on a new line if it can't fit on the current line, and then follows glyph wrapping.

### DIFF
--- a/benches/layout.rs
+++ b/benches/layout.rs
@@ -27,7 +27,7 @@ fn layout(c: &mut Criterion) {
             let mut run_on_text = |text: &str| {
                 buffer.lines.clear();
                 buffer.set_text(&mut fs, text, ct::Attrs::new(), *shape);
-                buffer.shape_until_scroll(&mut fs);
+                buffer.shape_until_scroll(&mut fs, false);
             };
 
             group.bench_function("small amount of text", |b| {
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -104,8 +104,10 @@ pub enum Wrap {
     None,
     /// Wraps at a glyph level
     Glyph,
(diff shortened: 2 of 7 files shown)


grovesNL:
Is there a way we could control whether to automatically fall back to glyph wrapping or not? I don't actually want to fallback to glyph wrapping for my use case, instead I intentionally clip the word if it can't fit onto the line

dtzxporter:
We could, I currently did it this way to match how the `word-break` CSS property behaves. I would assume it would make sense to still have this logic be 'default' with a way to optionally disable the fallback.

nicoburns:
> We could, I currently did it this way to match how the `word-break` CSS property behaves. I would assume it would make sense to still have this logic be 'default' with a way to optionally disable the fallback.

It seems that this PR matches `word-wrap: break-word` behaviour, whereas the existing behaviour matches the web default of `word-break: normal`. I don't have strong feelings on what the default is, but I feel like it definitely ought to be configurable.

dtzxporter:
@nicoburns @grovesNL I have added a new wrapping method that specifies glyph fallback should be used.

Now we can cover the most common `word-break` scenarios.

jackpot51:
Is this ready for review?

dtzxporter:
> Is this ready for review?

Yep!

jackpot51 (approved):

jackpot51:
I have set the default Buffer wrap to WordOrGlyph. I think it is a more sensible default as it prevents clipping. Buffer::set_wrap should of course be used if another wrap method is desired.

PR #226, opened 2024-02-02
Fix extra line insertion.

dtzxporter:
Another attempt at fixing #155 and another edge case where two overflowing words next to each other create a shadow run.

Passes all tests and adds a new test.

cc @grovesNL if you wouldn't mind testing this resolves your issue as well.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -1046,42 +1046,40 @@ impl ShapeLine {
                         } else {
                             // Wrap::Word, Wrap::WordOrGlyph
 
-                            // TODO: What if the previous span ended with whitespace and the next
-                            // span wraps a new line? Is that possible?
-                            //
-                            // TODO: This comment it outdated, the current word can be a
-                            // whitespace.
-                            //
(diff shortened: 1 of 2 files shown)


dtzxporter:
This is ready for review.

jackpot51 (approved):

src/shape.rs
```diff
@@ -1046,42 +1046,40 @@ impl ShapeLine {
                         } else {
                             // Wrap::Word, Wrap::WordOrGlyph
 
-                            // TODO: What if the previous span ended with whitespace and the next
-                            // span wraps a new line? Is that possible?
-                            //
-                            // TODO: This comment it outdated, the current word can be a
-                            // whitespace.
```
Imberflur, line 1053: Why were these notes removed? AFAICT the code change doesn't affect these?

alokedesai:
@jackpot51 Would you be open to cutting a new minor version of cosmic text that includes this fix? If not, could you provide more details on the typical release process and when we could expect this to be in a stable release? Thanks so much!

PR #228, opened 2024-02-06
Add serde support

StratusFearMe21:
This PR is actually a group of 3 PRs

-
-
-

Right now, I'm making a slideshow program which uses `cosmic-text` for shaping and layout. I'd love to be able to serialize my slideshows into a packed serde format, and in order to do that, I need the types in the libraries in these 3 PRs to implement the serde traits.


genusistimelord:
This might as well be closed, since they will not add serde support within fontdb or ttf-parser.

PR #230, opened 2024-02-12
Fix `no_std` build

### LINKED ISSUE #197: cosmic-text `0.11.0` is not `no_std` compatible
```
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv7em-none-eabi` target may not support the standard library
  = note: `std` is required by `unicode_properties` because it does not declare `#![no_std]`

   Compiling unicode-script v0.5.5
error: cannot find attribute `derive` in this scope
Error:   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/unicode-properties-0.1.0/src/tables.rs:44:7
   |
44 |     #[derive(Copy, Clone, Hash, Eq, PartialEq, Ord, PartialOrd, Debug)]
   |       ^^^^^^
   |
help: consider importing this attribute macro
   |
44 +     use core::prelude::rust_2024::derive;
   |
   ```
   
It seems like the `unicode-properties` doesn't support `no_std`. It probably makes sense for `cosmic-text` to add some CI target that tests against `no_std`.

It seems like there's already a fix on their side, but it's not released yet: https://github.com/unicode-rs/unicode-properties/pull/3

### PR DESCRIPTION
**Pull request by CryZe**
This fixes the `no_std` build and also makes sure it's tested properly in the CI workflow.

Resolves #197

### DIFF
--- a/src/glyph_cache.rs
+++ b/src/glyph_cache.rs
@@ -63,8 +63,9 @@ pub enum SubpixelBin {
 
 impl SubpixelBin {
     pub fn new(pos: f32) -> (i32, Self) {
-        let (fract, truncf) = libm::modff(pos);
-        let trunc = truncf as i32;
+        let trunc = pos as i32;
+        let fract = pos - trunc as f32;
+
         if pos.is_sign_negative() {
             if fract > -0.125 {
                 (trunc, Self::Zero)
--- a/src/math.rs
+++ b/src/math.rs
@@ -1,0 +1,14 @@
+#[cfg(not(feature = "std"))]
+pub use libm::{roundf, truncf};
+
+#[cfg(feature = "std")]
+#[inline]
+pub fn roundf(x: f32) -> f32 {
+    x.round()
+}
+
+#[cfg(feature = "std")]
+#[inline]
+pub fn truncf(x: f32) -> f32 {
+    x.trunc()
+}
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/math.rs [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+#[inline]
+pub fn roundf(x: f32) -> f32 {
+    x.round()
+}
+
+#[cfg(feature = "std")]
+#[inline]
+pub fn truncf(x: f32) -> f32 {
+    x.trunc()
+}
```
CryZe, line 14: I went ahead and introduced a math module. The idea is that `round`, `trunc` and co. are sometimes actual instructions, so using the simplistic software implementations in the `libm` crate is not ideal if you have `std` available.

src/glyph_cache.rs
```diff
@@ -63,8 +63,9 @@ pub enum SubpixelBin {
 
 impl SubpixelBin {
     pub fn new(pos: f32) -> (i32, Self) {
-        let (fract, truncf) = libm::modff(pos);
-        let trunc = truncf as i32;
+        let trunc = pos as i32;
+        let fract = pos - trunc as f32;
```
CryZe, line 67: Looking inside the implementation of `modff` reveals that it mostly just comes down to this here anyway as well as a bunch of special case handling (floats that are too large, too small, ...). I don't think those cases really matter here.

jackpot51 (approved):

PR #242, opened 2024-03-10
Faster mono fallback

MoSal:
* Cache codepoint support info for monospace fonts.
* Only try monospace fonts that support at least one requested script:
  * Codepoint support info is still collected for these fonts to guide
 the fallback order selection process.
  * A map of per-script monospace font-ids is pre-populated in font system
 to acquire lists of wanted font ids efficiently.
  * This gives an important performance boost when scripts are involved.
* Skip trying monospace fallbacks if the default font supports all codepoints.
  * Actually collect codepoint support info for the default font.
  * If it supports all codepoints, skip collecting that info from other
 monospace fonts.
  * Biggest performance improvement comes from this change, doubly so for users that only need the default font.

With all these changes, and with the relatively common single space word " ", monospace fallback process will go from:

* Checking for `space` codepoint support in all monospace fonts (except default font) using `slice::contains()`.
* Selecting default font first to try shaping with, which should work, unless it's a very unusual choice.

To the new process of:

* Checking for `space` codepoint support in default font.
  * `space` should be found quickly via a binary search in `FontCachedCodepointSupportInfo`.
* Selecting default font to try shaping with, which should work since we already checked that it does support `space`.

Should help with pop-os/cosmic-term#130.  
May or may not fully or partially solve pop-os/cosmic-term#155.

### DIFF
--- a/src/font/fallback/mod.rs
+++ b/src/font/fallback/mod.rs
@@ -2,6 +2,7 @@
 
 use alloc::collections::BTreeSet;
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use fontdb::Family;
 use unicode_script::Script;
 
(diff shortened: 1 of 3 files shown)


jackpot51 (approved):
Looks good, thanks!

PR #249, opened 2024-04-28
feat: concurrently load & parse fonts

Itsusinn:
This will reduce load time from 2376ms to 388ms when loading many huge fonts

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ default-features = false
 features = ["hardcoded-data"]
 
 [features]
-default = ["std", "swash", "fontconfig", "rayon"]
+default = ["std", "swash", "fontconfig"]
 fontconfig = ["fontdb/fontconfig", "std"]
 no_std = ["rustybuzz/libm", "hashbrown"]
 shape-run-cache = []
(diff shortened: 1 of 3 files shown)


Itsusinn:
@jackpot51 Could you please take a look?

Itsusinn:
https://docs.rs/rayon-core/1.12.1/rayon_core/index.html#global-fallback-when-threading-is-unsupported

I just find rayon supports `no_std`, maybe it's unnecessary to add `rayon` feature-gate?

CryZe:
That's not the same as `no_std`. If the crate doesn't have a `std` feature, then it doesn't support `no_std`.

Itsusinn:
> That's not the same as `no_std`. If the crate doesn't have a `std` feature (and otherwise uses `std` features), then it doesn't support `no_std`.

Understood

Cargo.toml
```diff
@@ -28,14 +28,15 @@ ttf-parser = { version = "0.20.0", default-features = false }
...
+rayon = { version = "1", optional = true }
 
 [dependencies.unicode-bidi]
 version = "0.3.13"
 default-features = false
 features = ["hardcoded-data"]
 
 [features]
-default = ["std", "swash", "fontconfig"]
+default = ["std", "swash", "fontconfig", "rayon"]
```
genusistimelord, line 39: I would either Place Rayon under the Std feature, OR not add it to defaults and add the std as its own requirement.

Personally, adding it under std seems to be a decent choice here if you want it defaulted.

genusistimelord (changes requested):
Everything But the feature itself, being added as a default, looks good.  Once that is fixed I think this would be ready to merge as I see no issues coming from this.  @jackpot51

Cargo.toml
jackpot51: I agree

genusistimelord (approved):
looks much better thank you

jackpot51 (approved):

Issue #227, opened 2024-02-06
Emoji issues


snaggen:
Using Cosmic-Term and looking at the file https://github.com/pop-os/cosmic-text/blob/main/sample/emoji.txt with both fonts DejaVu Sans and Noto Color Emoji installed show two issues. 

![image](https://github.com/pop-os/cosmic-text/assets/6420639/33d763c0-1251-473b-872b-e63c7b60e6c6)

1. The monochrome font DejaVu Sans is the preferred font, but Noto Color Emoji is clearly what a user would prefer. That is only used for glyphs missing from DejaVu Sans.
2. 263A FE0F , 263A, 1F642 gets a to narrow width

I think there is some color property for a font that can be used to prioritize a color font over monochrome fonts, that would solve issue 1.

Version info:
Cosmic-Term master from 2024-02-06
Cosmic Term uses cosmic-text git+https://github.com/pop-os/cosmic-text.git#be471833c9ee563866f07593354a41dd40761a18"
DejaVu Sans: dejavu-sans-mono-fonts-2.37-20.fc39.noarch
Noto Color Emoji: google-noto-color-emoji-fonts-20220916-7.fc39.noarch

PR #256, opened 2024-06-04
Promote Noto Color Emoji fallback

apricotbucket28:
Helps with https://github.com/pop-os/cosmic-text/issues/227

The DejaVu Sans font includes some emoji glyphs, so I moved it after Noto Color Emoji in the fallback  order.

For some reason the smiling face emoji still isn't correct, but rendering of other emojis has been fixed.
image


jackpot51 (approved):

dharmik2319:
@jackpot51 Why might those specific emojis not be in color? I am currently facing this problem

PR #257, opened 2024-06-06
Add metrics to attributes

### LINKED ISSUE #64: Spans of font sizes

jackpot51:
Rewrite of #235, fixes #64

### DIFF
--- a/src/attrs.rs
+++ b/src/attrs.rs
@@ -105,8 +105,8 @@ impl FamilyOwned {
 //TODO: what are the edge cases of this?
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct CacheMetrics {
-    pub font_size_bits: u32,
-    pub line_height_bits: u32,
+    font_size_bits: u32,
+    line_height_bits: u32,
 }
 
 impl From<Metrics> for CacheMetrics {
--- a/src/attrs.rs
+++ b/src/attrs.rs
@@ -8,7 +8,7 @@ use alloc::{
 use core::ops::Range;
 use rangemap::RangeMap;
 
-use crate::CacheKeyFlags;
+use crate::{CacheKeyFlags, Metrics};
 
 pub use fontdb::{Family, Stretch, Style, Weight};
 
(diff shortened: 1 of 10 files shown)


src/attrs.rs
```diff
@@ -101,6 +101,32 @@ impl FamilyOwned {
     }
 }
 
+/// Metrics, but implementing Eq and Hash using u32 representation of f32
+//TODO: what are the edge cases of this?
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct CacheMetrics {
+    pub font_size_bits: u32,
```
alice-i-cecile, line 108: I would be very reluctant to expose these fields directly. Getters / setters seem like the way to go here.

src/shape.rs
```diff
@@ -513,15 +518,17 @@ impl ShapeGlyph {
             cache_key_flags: self.cache_key_flags,
         }
     }
+
+    pub fn width(&self, font_size: f32) -> f32 {
```
alice-i-cecile, line 522: Doc strings would be helpful here. Units are the main question.

src/attrs.rs
jackpot51: I think just the into/from implementations for Metrics are enough.

nicoburns (changes requested):
Haven't had chance to review the code yet, but there is bug when creating new lines:

Screenshot 2024-06-06 at 23 41 06

(newly created lines end up with the "font size 64" line-height even though there is no text on that line with that height)

jackpot51:
Right now it takes the maximum of either the buffer line height or the glyph line height for the whole run. I could probably only reference the glyph line height but I didn't want to run into empty lines being 0 height.

jackpot51:
@nicoburns should be resolved in the latest commit.

alice-i-cecile (approved):
This seems reasonable from a code quality perspective, and I can follow the logic :)

jackpot51:
This is working well enough for me that I will merge. Any issues I missed can be reported against master.

PR #261, opened 2024-06-08
Add IME preedit ranges

Riateche:
Added `SetPreedit` action that inserts text with a special `is_preedit` flag in the attributes and clears any previous preedit. This is the most straightforward way to show a preedit inline, so that it works naturally with shaping and word wrapping. The downside is that the buffer's text is not the same as the document text anymore. For example, if the user wants to retrieve the text from the buffer, they should delete the part corresponding to the preedit (or clear the preedit). To help with this, I added `Buffer::text_without_preedit()` and `BufferLine::text_without_preedit()`.

I've added IME support to the `editor` example to demonstrate the correct handling of IME.


Riateche:
Rebased onto main and fixed a CI issue with no_std. Please take a look.

PR #262, opened 2024-06-08
Add Edit::cursor_position

Riateche:
Added `Edit::cursor_position()` method to retrieve the coordinates of the visual cursor. This is required to set the IME position, so that the input method window is displayed close to the cursor. The current implementation simply reuses the `Editor::draw()` code (moved out to a function). This can probably be done more efficiently.

### DIFF
--- a/src/edit/editor.rs
+++ b/src/edit/editor.rs
@@ -12,7 +12,7 @@ use unicode_segmentation::UnicodeSegmentation;
 use crate::Color;
 use crate::{
     Action, Attrs, AttrsList, BorrowedWithFontSystem, BufferLine, BufferRef, Change, ChangeItem,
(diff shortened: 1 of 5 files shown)


src/edit/editor.rs
```diff
@@ -28,6 +28,72 @@ pub struct Editor<'buffer> {
     change: Option<Change>,
 }
 
+fn cursor_glyph_opt(cursor: &Cursor, run: &LayoutRun) -> Option<(usize, f32)> {
```
jackpot51, line 31: Should this be a method on LayoutRun or Cursor?

jackpot51 (approved):
The internal API is a bit strange in my opinion, but the new Edit::cursor_position function seems fine to me.

PR #263, opened 2024-06-08
Hide cursor based on selection, preedit, or config

Riateche:
1. Cursor is now hidden when there is a selection.
2. Cursor is now hidden when there is an IME preedit that specifies that there should be no cursor.
3. Added `Edit::set_cursor_hidden()` to hide cursor manually, which can be used to implement cursor blinking, and to hide cursor when the input is unfocused or read only.

While I generally prefer "positive" boolean flags (e.g. `set_cursor_visible()` would generally be better), in this case the logic looks simpler when we use the `hidden` flag: `cursor_hidden = has_selection || has_preedit_without_cursor || set_cursor_hidden`.

### DIFF
--- a/src/edit/editor.rs
+++ b/src/edit/editor.rs
@@ -25,6 +25,11 @@ pub struct Editor<'buffer> {
     cursor_moved: bool,
     auto_indent: bool,
     change: Option<Change>,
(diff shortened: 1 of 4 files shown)


Riateche:
Rebased onto main. Please take a look.

PR #265, opened 2024-06-08
Add select flag for motion and click actions

Riateche:
Added `select` flag to `Action::Motion` and `Action::Click` to implement hotkeys with `Shift` modifier (e.g. `Shift + ArrowRight` to create or extend the selection) and fix hotkeys without `Shift` (e.g. `ArrowRight` previously didn't clear existing selection). 

Fixed `Previous` and `Next` behavior when there is a selection. They now clear selection and put the cursor to the start and end of the selection, respectively.

Implemented `select` flag for examples.

### DIFF
--- a/examples/editor-test/src/main.rs
+++ b/examples/editor-test/src/main.rs
@@ -116,7 +116,10 @@ fn main() {
             // Test delete of EGC
             {
                 let cursor = editor.cursor();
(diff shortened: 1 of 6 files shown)


jackpot51 (commented):
I will have to test how this works with ViEditor before approving

Riateche:
Rebased onto main. Please take a look.

PR #284, opened 2024-07-10
implement Clone for Editor

### LINKED ISSUE #283: Editor Clone
Is there a reason why [`Editor`](https://docs.rs/cosmic-text/latest/cosmic_text/struct.Editor.html) is not Clone?

I think it would be usefull to some applications to have the ability to clone a Editor. My app for example has issues with this (can not use editor because of it).

JL710:
This PR closes #283.

The only mentionable to this PR is that BufferRef::Borrowed is converted to BufferRef::Owned when cloning. 
```rs
impl<'buffer> Clone for BufferRef<'buffer> {
    fn clone(&self) -> Self {
        match self {
            Self::Owned(buffer) => Self::Owned(buffer.clone()),
            Self::Borrowed(buffer) => Self::Owned((*buffer).to_owned()),
            Self::Arc(buffer) => Self::Arc(buffer.clone()),
        }
    }
}
```

### DIFF
--- a/src/edit/editor.rs
+++ b/src/edit/editor.rs
@@ -16,7 +16,7 @@ use crate::{
 };
 
 /// A wrapper of [`Buffer`] for easy editing
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Editor<'buffer> {
     buffer_ref: BufferRef<'buffer>,
     cursor: Cursor,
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jackpot51:
It looks like this has an issue when building for no_std, can you fix that?

JL710:
> It looks like this has an issue when building for no_std, can you fix that?

I have never worked with no_std builds. How to I build that?

JL710:
> It looks like this has an issue when building for no_std, can you fix that?

I fixed no_std in a new commit.

JL710:
@jackpot51

Do I need to do something else for this PR to get merged? It seems one workflow is failing, but I am not familiar with the reasons.

jackpot51:
I think it is fine, I will merge.

PR #294, opened 2024-08-01
update fontdb, rustybuzz and ttf-parser

MaxVerevkin:
Currently there is a dependency on both `ttf-parser v0.20.0` and `ttf-parser v0.21.0`, so I decided to update everything to the latest `ttf-parser v0.24.0`.


jackpot51 (approved):

jackpot51:
This unfortunately was not working for iced/libcosmic because usvg is still relying on an older fontdb version.

PR #298, opened 2024-08-10
improve the performance of trimming shape run cache

dzhou121:
When shape run cache was trimmed with a large `keep_ages`, it's expensive to scan the whole cache.

This PR is trying to save the cache keys to a Vec of age registries, and when doing trim, pop the `Vec` according to `keep_ages` and those are the keys that needed to be removed from the cache.

### DIFF
--- a/src/shape_run_cache.rs
+++ b/src/shape_run_cache.rs
@@ -2,7 +2,7 @@
 use alloc::{string::String, vec::Vec};
 use core::ops::Range;
 
-use crate::{AttrsOwned, HashMap, ShapeGlyph};
+use crate::{AttrsOwned, HashMap, HashSet, ShapeGlyph};
 
 /// Key for caching shape runs.
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


src/shape_run_cache.rs
```diff
@@ -13,32 +13,68 @@ pub struct ShapeRunKey {
 }
 
 /// A helper structure for caching shape runs.
-#[derive(Clone, Default)]
+#[derive(Clone)]
 pub struct ShapeRunCache {
     age: u64,
     cache: HashMap<ShapeRunKey, (u64, Vec<ShapeGlyph>)>,
+    age_regitries: Vec<HashSet<ShapeRunKey>>,
```
jackpot51, line 20: This should be age_registries

jackpot51:
Sorry for the delay in reviewing. I have one item I would like fixed.

src/shape_run_cache.rs
dzhou121: should be fixed now

jackpot51 (dismissed):

jackpot51:
Looks like Rust/build CI is failing

src/shape_run_cache.rs
```diff
@@ -13,13 +13,27 @@ pub struct ShapeRunKey {
...
+impl Default for ShapeRunCache {
+    #[allow(clippy::vec_init_then_push)]
+    fn default() -> Self {
+        let mut age_registries = Vec::new();
+        age_registries.push(HashSet::default());
+
+        Self {
+            age: 0,
+            cache: Default::default(),
+            age_registries,
```
jackpot51, line 32: You could juts do `age_registries: vec![HashSet::default()],`

jackpot51 (approved):

jackpot51:
This looks good to me. Given that @UkoeHB also provided feedback, I'd like a re-review from them as well.

PR #302, opened 2024-08-31
Small optimizations

UkoeHB:
- Use `SmolStr` instead of `String` in `FamilyOwned`. Families are almost always <= 23 bytes. Note that `SmolStr` is `no-std` when default features are off.
- Add `AttrsList::spans_iter` to avoid an unnecessary allocation in `Buffer::append`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ rangemap = "1.4.0"
 rustc-hash = { version = "1.1.0", default-features = false }
 rustybuzz = { version = "0.14", default-features = false, features = ["libm"] }
 self_cell = "1.0.1"
+smol_str = { version = "0.2.2", default-features = false }
 swash = { version = "0.1.17", optional = true }
 syntect = { version = "5.1.0", optional = true }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -21,6 +21,7 @@ rangemap = "1.4.0"
 rustc-hash = { version = "1.1.0", default-features = false }
 rustybuzz = { version = "0.14", default-features = false, features = ["libm"] }
 self_cell = "1.0.1"
+smol_str = { version = "0.2.2", default-features = false }
```
grovesNL, line 24: For what it's worth, `smol_str` doesn't seem to be recommended anymore (see the comment by the original author in https://www.reddit.com/r/rust/comments/117ksvr/comment/j9eh35d/)

There's also a comparison of small string crates at https://github.com/rosetta-rs/string-rosetta-rs?tab=readme-ov-file (`smol_str` was removed from the table in favor of ecow)

jackpot51 (approved):

Issue #301, opened 2024-08-31
Efficiency of Buffer::set_rich_text


UkoeHB:
Currently `Buffer::set_rich_text` uses a lot of allocations, even when reconstructing a buffer. Ideally most or all of these allocations could be amortized using cached buffers and reusing allocations within the Buffer more aggressively. A full solution would require passing buffers and buffer sources to the method, so maybe a new method could be added for use-cases where full optimization is important (e.g. `bevy`).

As an aside, the `Family::Name` variant contains a `String` but it would make more sense as a `SmolStr`.

PR #303, opened 2024-08-31
Big optimizations for Buffer::set_rich_text

UkoeHB:
## Problem

Currently when `Buffer::set_rich_text` is called, all existing allocations in the buffer are simply discarded. This amounts to a *lot* of allocations for complex text.

## Solution

Reuse existing allocations as much as possible when rebuilding a buffer.

Note that the changes here try to line up buffer reuse so it all maps back into the same positions in the buffer if possible. This way if, for example, you are just animating text color, the buffer won't actually require any extra allocations after the first time it's constructed (and won't over-allocate).

I did not originally intend for this PR to get so big, but diving into the code was a rabbit hole of allocations to squash.

Closes https://github.com/pop-os/cosmic-text/issues/301

## Testing

I did not do any perf testing.

## Follow-up

Additional areas for improvement:
- `Buffer::set_text` can also be optimized in a similar fashion.
- Use `SmallVec` instead of `Vec` where it makes sense. This should somewhat reduce the cost to make new buffers.
- There is one allocation at the top of `set_rich_text` that I didn't try to solve yet.
- Internally, `ShapeLine` is essentially a `Vec<Vec<Vec<ShapeGlyph>>>`. This can be flattened into two vecs: a vec of indices + metadata and a vec of ShapeGlyphs. Or even one vec with the metadata in-line.
- Internally, `BufferLine` caches a `Vec<LayoutLine>`, which is essentially `Vec<Vec<LayoutGlyph>>`. This could also be flattened.
- `Buffer` caches a `ShapeBuffer` but it would be better to have this in `FontSystem` or elsewhere for global reuse.

### DIFF
--- a/src/buffer_line.rs
+++ b/src/buffer_line.rs
@@ -1,5 +1,6 @@
 #[cfg(not(feature = "std"))]
 use alloc::{string::String, vec::Vec};
+use core::mem;
 
 use crate::{
     Align, Attrs, AttrsList, Cached, FontSystem, LayoutLine, LineEnding, ShapeBuffer, ShapeLine,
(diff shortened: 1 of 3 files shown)


jackpot51 (approved):

jackpot51:
The Rust/build CI will need to pass

PR #306, opened 2024-09-01
add the convenient API for setting tab width

### LINKED ISSUE #305: Can I use `set_tab_width` without passing a `&mut FontSystem`.
This code currently doesn't work
```rs
    let mut buffer = Buffer::new(&mut font_system, metrics);
    
    // This does not work with set_tab_width for some reason.
    let mut buffer = buffer.borrow_with(&mut font_system);
    
    buffer.set_tab_width(args.tab_width);
```

I would have to change every method call on the buffer so it takes the font system.
```rs
    buffer.set_tab_width(&mut font_system, args.tab_width);
```

Is there any barrier to adding this somewhere about here? 
https://github.com/pop-os/cosmic-text/blob/main/src/buffer.rs#L1326-L1431

wildwestrom:
Fixes #305.

By all means let me know if there are any problems, I'll fix them as soon as possible.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1379,6 +1379,11 @@ impl<'a> BorrowedWithFontSystem<'a, Buffer> {
             .set_metrics_and_size(self.font_system, metrics, width_opt, height_opt);
     }
 
+    /// Set tab width (number of spaces between tab stops)
+    pub fn set_tab_width(&mut self, tab_width: u16) {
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

wildwestrom:
Thanks @jackpot51! When can I expect this to be on crates.io?

jackpot51:
There are a few other PRs that are waiting for merge. Hopefully I can do a new release tomorrow.

PR #311, opened 2024-09-02
Optimize for single-line single-span single-word text

UkoeHB:
Use `SmallVec` with the assumption that a large proportion of text is single-line, and possible single-span and single-word.

I tried using `SmallVec` for glyphs but noticed a perf regression in one unit test.


grovesNL:
It would be nice if this could be configurable somehow. This would be a regression my application - almost all of my buffers have more than one word and span more than one line.

PR #323, opened 2024-10-24
Support subpixel rendering

jackpot51:
Fixes #140 

Draft until it is configurable

### DIFF
--- a/examples/editor-test/src/main.rs
+++ b/examples/editor-test/src/main.rs
@@ -31,7 +31,8 @@ fn redraw(
             cursor_color,
             selection_color,
             selected_text_color,
-            |x, y, w, h, color| {
+            |x, y, w, h, color, subpixel_mask| {
+                assert!(subpixel_mask.is_none());
                 window.rect(x, y, w, h, orbclient::Color { data: color.0 });
             },
         );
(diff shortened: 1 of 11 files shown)


charliechensz:
## Data point: low-PPI users are hitting this in practice, and a working patch needs consumer-side support too

Adding a data point from a real 140 PPI deployment (Pop!_OS 24.04, cosmic-term 1.8.0, 31.5″ 4K LCD):

- **Impact**: grayscale-only AA (`Format::Alpha` hardcoded in `swash.rs`) is clearly visible below ~150 PPI on stroke-heavy text (Latin `o`/`e`/`s` and CJK alike). On a 259 PPI panel, same version/config looks fine — it's purely a PPI × rasterizer interaction.
- **User-facing config does not help**: fontconfig `rgba`/`hintstyle`, `wl_output` subpixel order, font family, and font size were all verified ineffective (cosmic-text reads neither fontconfig rendering params nor output subpixel geometry). Details in cosmic-term #363.
- **For the fix itself**: PR #532 (cosmic-text subpixel masks) works at the library level — verified `SubpixelMask` output with correct BGR stripe order. But cosmic-term's renderer (glyphon) currently maps `SubpixelMask => ContentType::Mask // "Not implemented yet"`, so end-to-end output is blurry. A production fix needs the mask producer (cosmic-text) **and** a consumer (glyphon) that can composite subpixel masks (RGBA per-stripe coverage + subpixel blending).

Happy to test a combined change on the 140 PPI panel.

Issue #327, opened 2024-11-16
Inconsistent emoji rendering


richardhozak:
Hello, while trying to update cosmic-text from version v0.7-ish (commit https://github.com/pop-os/cosmic-text/commit/e00109d77f06d5a2e3057865eda3f530bc40a046) to latest version v0.12.1 I've noticed there is a regression in rendering emojis.

The latest version renders only some emojis in color while rendering other in black/white version.

Here is a screenshot how it is rendered with 0.12.1
![Screenshot from 2024-11-16 20-05-13](https://github.com/user-attachments/assets/355d1e73-32f4-4c6c-b715-65a721723cf5)

And here is a screenshot on how it was rendered with https://github.com/pop-os/cosmic-text/commit/e00109d77f06d5a2e3057865eda3f530bc40a046
![image](https://github.com/user-attachments/assets/d2423855-e765-4107-b02b-1a53f430e908)

Notice the different emojis. The text also looks different but not incorrect.

I've also tested this with v0.11, v0.10, v0.9 and v0.8 and they all look the same as v0.12.1.
I've also tested with release v0.7, that also looks the same as v0.12.1. Then I tried downgrading to 0.6 and the emojis render correctly, so it must be some change between 0.6 and 0.7.

richardhozak:
So I've checked all the commits between https://github.com/pop-os/cosmic-text/compare/0.6.0...0.7.0 and found the regression which is commit https://github.com/pop-os/cosmic-text/commit/66288ab2da9e5e9e74ede1ed081e581058ab4cca Noto Color Emoji should probably be higher in the fallback list. Will try to implement fix...

dharmik2319:
@richardhozak Any updates? I am facing the same problem

richardhozak:
The issue is still present unfortunately.

The fix/revert proposed in https://github.com/pop-os/cosmic-text/pull/328 works for me but the best solution seems yet to be found, you can check the discussion on the PR.

modelmiser:
I'm hitting this in an iced app (some emoji render colored, others monochrome) and would like to work on a fix.

Looking at the code: the issue is that `common_fallback()` on Unix puts Noto Sans / DejaVu Sans / Noto Sans Symbols2 before Noto Color Emoji, and these fonts contain monochrome outline glyphs for many emoji codepoints, so `shape_fallback()` never reports them as missing. The fallback chain stops before reaching Noto Color Emoji.

[...]

PR #328, opened 2024-11-16
Revert "Fix `Noto Color Emoji` font fallback order for Unix"

richardhozak:
This reverts commit 66288ab2da9e5e9e74ede1ed081e581058ab4cca.

Note that putting "Noto Color Emoji" anywhere lower (even just below "DejaVu Sans") creates issue described in https://github.com/pop-os/cosmic-text/issues/327

Putting it above "Noto Sans" breaks text rendering completely.

With the commit reverted I cannot reproduce the original issue https://github.com/pop-os/cosmic-text/pull/68 the commit 66288ab2da9e5e9e74ede1ed081e581058ab4cca was supposed to fix.

### DIFF
--- a/src/font/fallback/unix.rs
+++ b/src/font/fallback/unix.rs
@@ -8,6 +8,8 @@ pub fn common_fallback() -> &'static [&'static str] {
     &[
         /* Sans-serif fallbacks */
         "Noto Sans",
+        /* Emoji fallbacks*/
+        "Noto Color Emoji",
         /* More sans-serif fallbacks */
         "DejaVu Sans",
         "FreeSans",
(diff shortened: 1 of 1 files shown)


richardhozak:
For posterity here are some screenshots...

With this PR:
Screenshot from 2024-11-16 21-50-52

When "Noto Color Emoji" is first in the array (above "Noto Sans"), text rendering breaks completely, but emojis work:
Screenshot from 2024-11-16 21-50-00

When "Noto Color Emoji" is just below "DejaVu Sans", text rendering does not display emojis properly:
image

jackpot51:
Perhaps @hecrj can check if this works okay on their system?

hecrj:
@jackpot51 Just tried it and I have the same issues:

image

hecrj:
I believe a proper fix would be to have a separate emoji fallback list and have some kind of emoji detection during font selection.

PR #369, opened 2025-03-15
Configurable font fallback lists

### LINKED ISSUE #126: Allow setting the fallback fonts

tigregalis:
Fixes #126

A configurable font fallback list can be set by the user during construction of the `FontSystem`.

Commit #0fb1cadb essentially preserves the exact existing behaviour. The further commits after that allow for non-static fallback lists (i.e. not known at compile-time), however these names are still `&'static str`.

Changes to the public API:

1. A new trait `Fallback`.
1. A new constructor `FontSystem::new_with_locale_and_db_and_fallback`, allowing you to pass in a type that implements `Fallback`.
1. A new struct `PlatformFallback` that implements `Fallback`, which encapsulates the existing platform-specific fallback lists in `src/font/fallback/windows.rs` etc.

At `FontSystem` construction, the fallback lists are collected into a private `Fallbacks` struct, which is done to limit the number of times the `Box<dyn Fallback>` is accessed. Each time the `FontFallbackIter` is constructed, the `Fallbacks` struct is extended with the relevant script-specific fallback lists.

In my tests, there is a small regression in performance in `Wrap(None, Simple)`, but no regression in all the other benchmarks, and even supposed performance improvement in some of the more complex text benchmarks but I can't see how this could be the case.

# Possible future work

1. The names of the fonts are currently `&'static str`, but this could be expanded to `&str`, for more run-time configurability.
1. It is possible to eliminate the `Box<dyn Fallback>`, if the `Fallback` trait adds a `supported_scripts` method, then at `FontSystem` construction time, enumerates the supported scripts and their fallback lists, also eliminating the need to extend the `Fallbacks` struct each time the `FontFallbackIter` is constructed.
1. The trait methods could return iterators instead of slices, for more flexibility.
1. A different storage/lookup mechanism for the `script_fallbacks` (instead of the current `HashMap`) could theoretically be worthwhile for performance.
1. Going in the other direction, there could be more dynamic/bespoke behaviour rather than these effectively flat lists of fallback fonts.

### DIFF
--- a/src/font/fallback/macos.rs
+++ b/src/font/fallback/macos.rs
@@ -2,8 +2,31 @@
 
 use unicode_script::Script;
 
+use super::Fallback;
+
+/// A platform-specific font fallback list, for MacOS.
+pub struct PlatformFallback;
+
+impl Fallback for PlatformFallback {
+    fn common_fallback(&self) -> &'static [&'static str] {
+        common_fallback()
+    }
+
+    fn forbidden_fallback(&self) -> &'static [&'static str] {
+        forbidden_fallback()
+    }
+
+    fn script_fallback(
+        &self,
+        script: unicode_script::Script,
+        locale: &str,
+    ) -> &'static [&'static str] {
+        script_fallback(script, locale)
+    }
+}
+
 // Fallbacks to use after any script specific fallbacks
-pub fn common_fallback() -> &'static [&'static str] {
+fn common_fallback() -> &'static [&'static str] {
     &[
         ".SF NS",
         "Menlo",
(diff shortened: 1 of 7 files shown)


bes:
I'm building a cross-platform application (web, iOS, Android, etc.) and this would be great because on web specifically I want:

* Default font to be `Inter`, but inter doesn't support e.g. Arabic, and many other scripts
* So I want to load a fallback font for the other scripts, e.g. `Noto Sans`

StewartCanva:
This is fantastic. This is exactly what we were looking for with our project.

jackpot51 (approved):

PR #371, opened 2025-03-21
Add support for kerning and ligature font features

StewartCanva:
Right now cosmic text is not passing any font feature information into rustybuzz. This is fine for most cases, but in our use case we want to be able to control whether kerning and ligature is applied or not on text (whether specified in the font file or using the fallback mechanism in rustbuzz) . Rustybuzz will without intervention, indiscriminately apply kerning and ligatures for all text.

This would also open up the pathway for more font feature support within the future.

On a side note. We're trying to use this crate at Canva and there's a few other additions we would love to add to cosmic text in order to increase its functionality. Would love to get the conversation rolling in terms of that to see if we can invest our development time into it.

### DIFF
--- a/src/attrs.rs
+++ b/src/attrs.rs
@@ -5,6 +5,8 @@ use alloc::vec::Vec;
 use core::ops::Range;
 use rangemap::RangeMap;
 use smol_str::SmolStr;
+use std::collections::HashMap;
+use std::rc::Rc;
 
 use crate::{CacheKeyFlags, Metrics};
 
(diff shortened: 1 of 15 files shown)


src/attrs.rs
```diff
@@ -125,6 +125,34 @@ impl From<CacheMetrics> for Metrics {
     }
 }
 
+/// Font features for controlling typographic behavior
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct FontFeatures {
```
jackpot51, line 130: This looks very good! I think it would be nice to document the font feature name for each of these, like `liga`

src/attrs.rs
```diff
@@ -204,6 +234,12 @@ impl<'a> Attrs<'a> {
...
     }
 
+    /// Set [FontFeatures]
+    pub fn font_features(mut self, font_features: FontFeatures) -> Self {
+        self.font_features = font_features;
+        self
+    }
+
     /// Check if font matches
     pub fn matches(&self, face: &fontdb::FaceInfo) -> bool {
```
jackpot51, line 244: Should the matches function be checking font features? This is used to determine what is grouped into a run for shaping, so it seems important.

jackpot51:
@StewartCanva I'm interested in improving cosmic-text for all use cases, your contributions are welcome.

src/attrs.rs
```diff
@@ -137,6 +201,7 @@ pub struct Attrs<'a> {
     pub metadata: usize,
     pub cache_key_flags: CacheKeyFlags,
     pub metrics_opt: Option<CacheMetrics>,
+    pub font_features: FontFeatures,
```
StewartCanva, line 204: @jackpot51 I realised my first proposal, while it would work for kerning and ligature support, would not scale further for future font feature support. I was adding a `bool` or `u32` per font feature with the previous one, and at over `120` different font features that will add a massive memory overhead. 😅

This version I guess is more similar to the rustybuzz interface and scales better. The tradeoff is that `Copy` cannot be applied on `Attrs` which does have slight implications for consumers of the crate (namely the use of `.clone()` depending on how a consumer is associating a span with an attr). There's also some other places throughout the crate that need to be updated because of this. Which I'll do if you think this approach is fine. I tried to avoid it, but there doesn't seem to be a better way.

jackpot51: That seems fine to me

jackpot51:
CI is having issues, can you take a look?

jackpot51:
it looks like some more places need to clone Attrs instead of trying to copy

jackpot51 (approved):

PR #373, opened 2025-03-28
Add letter spacing support

StewartCanva:
Add in support for letter spacing equivalent to https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing , however it's defined only in em rather than being able to supply both px or em. Letter spacing adds a uniform additional spacing between glyphs. So it does not replace kerning or other spacing logic, but rather augments it. Unless the user specifies a non-zero letter spacing value the current logic remains the same.

Resolves a request for it under https://github.com/pop-os/cosmic-text/discussions/368

Screenshot 2025-03-28 at 4 40 56 pm

### DIFF
--- a/src/attrs.rs
+++ b/src/attrs.rs
@@ -5,6 +5,7 @@ use alloc::vec::Vec;
 use core::ops::Range;
 use rangemap::RangeMap;
 use smol_str::SmolStr;
+use std::hash::{Hash, Hasher};
 
 use crate::{CacheKeyFlags, Metrics};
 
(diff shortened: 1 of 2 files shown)


shiro:
Nice, look amazing!
Just curious, on an implementation level, would it be difficult to support both `px` and `em` units?

StewartCanva:
> Nice, look amazing! Just curious, on an implementation level, would it be difficult to support both `px` and `em` units?

Thanks. Implementation wise it's not hard to do. I had it originally for both. The only issue was the way I had it setup  required adding a new `f32` field per `ShapeGlyph` and then you'd calculate it dynamically under `width()` in order to support px values (I guess not the worse thing ever but it'll likely be unused 99% of the time). There's probably other ways to do it but I didn't spend a whole lot of time on it. I figured since most users wouldn't use this then it probably wasn't worth the effort. And while not the best dev experience it's possible to convert from em to px on the client side.

test3211234:
@StewartCanva Problem, the connecting parts of letters in cursive fonts are separate and are spaced from the characters they are attached to. They should be attached to the preceding character.

{3C2E67A7-F5B6-4BAC-B40A-ECFBC2E11161}

PR #375, opened 2025-04-04
Store font as peniko::Font behind feature flag

taj-p:
## Intent

Behind a feature flag, allow fonts to be stored and retrieved as `peniko::Font`.

## Background

I work with @StewartCanva on a part of our renderer that interfaces with [Vello](https://github.com/linebender/vello). To allow Cosmic Text to interface with the Linebender stack, it's preferable to store font data using [`peniko::Font`](https://github.com/linebender/peniko/blob/615caa3fdb0df9e8396e4efe4ab89fe8e1a872a7/src/font.rs#L8). Otherwise, we'd likely need to doubly allocate font data, which is something we'd very much like to avoid.

### DIFF
--- a/src/font/mod.rs
+++ b/src/font/mod.rs
@@ -91,8 +91,8 @@ impl Font {
     }
 
     #[cfg(feature = "peniko")]
-    pub fn peniko(&self) -> &peniko::Font {
-        &self.data
+    pub fn as_peniko(&self) -> peniko::Font {
+        self.data.clone()
     }
 
     #[cfg(feature = "swash")]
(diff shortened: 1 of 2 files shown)


src/font/mod.rs
```diff
@@ -73,13 +79,25 @@ impl Font {
...
+        }
     }
 
     pub fn rustybuzz(&self) -> &RustybuzzFace<'_> {
         self.rustybuzz.borrow_dependent()
     }
 
+    #[cfg(feature = "peniko")]
+    pub fn as_peniko(&self) -> PenikoFont {
+        self.data.clone()
```
taj-p, line 98: This clones an underlying `Arc` in [Blob](https://github.com/linebender/peniko/blob/615caa3fdb0df9e8396e4efe4ab89fe8e1a872a7/src/blob.rs#L13)

jackpot51 (approved):

jackpot51:
@taj-p Released as 0.14.1

PR #392, opened 2025-06-11
Add check for ligature patterns

### LINKED ISSUE #378: Some ligatures are not being applied
Not all ligatures are functional for certain fonts, even when explicitly enabling them.

## How to reproduce
Font: [Maple Mono NF](https://github.com/subframe7536/maple-font)
Features: `calt`

Edit an example from the project to render the following string with the given font and features:
```
/* /** ++ +++ <-> <|> ||> |> --> -> != !== =!= :?> ||= |= =/= <-< |=> -~ </> /> <!---->
```

All of the ligatures present in the string are (partially) not being applied, or incorrectly.
Image

Whereas all of the other ones work fine (e.g. `=>`, `==`).

For reference, this is what the same string should look like:

Image

EDIT: I have confirmed the following fonts also experience this issue:
- Iosevka
- Fira Code

jordandelbar:
I recently encountered the [ligature issue in Zed](https://github.com/zed-industries/zed/issues/12176) and found #378. 
This PR adds a check to ensure that ligature patterns are not broken.

disclaimer: font rendering and font shaping are not within my domain of competence and I shamelessly vibe coded it + verify that the tests + clippy & examples pass but I am not in capacity to ensure that it does not have other impacts (performance wise, etc) or if this is the correct way to fix this.

Closes #378

In zed 

before:
Screenshot From 2025-06-11 16-29-50

after (built with my branch as source for cosmic-text):
Screenshot From 2025-06-11 16-42-52

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -737,7 +737,10 @@ impl ShapeSpan {
         }
 
         let mut start_word = 0;
-        for (end_lb, _) in unicode_linebreak::linebreaks(span) {
+        let linebreaks: Vec<_> = unicode_linebreak::linebreaks(span).collect();
+        let merged_breaks = merge_ligature_breaks(&linebreaks, span);
+
+        for (end_lb, _) in merged_breaks {
             let mut start_lb = end_lb;
             for (i, c) in span[start_word..end_lb].char_indices().rev() {
                 // TODO: Not all whitespace characters are linebreakable, e.g. 00A0 (No-break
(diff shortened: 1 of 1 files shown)


jackpot51:
This may be okay, but I need to do testing, and there could be a way to fix this without hardcoding ligatures.

peppidesu:
I don't know if there is a way to retrieve a list of ligatures of a font. This would still leave another issue though: ligatures should be broken when necessary, which this solution prevents from happening.

jordandelbar:
I’m going to close this since, as @peppidesu [pointed out](https://github.com/zed-industries/zed/issues/12176#issuecomment-3097593912), it doesn’t actually solve the underlying issue. My goal was mainly to surface where the ligature rendering problem comes from in Zed. This isn’t really my area of expertise, so I’m not able to provide a proper fix. I had hoped the PR would spark some traction, but it didn’t.

jackpot51:
Unfortunately I'm not fully understanding how they are breaking in the first place. If that could be narrowed down, it would help with finding a solution.

PR #395, opened 2025-06-20
Fix wrap_word_fallback test: avoid temporary borrow, improve clarity

KushalMeghani1644:
### ✅ What this PR does:
- Replaces `.unwrap()` with `.expect(...)` for clearer error output.
- Ensures correct `Option<f32>` usage for `set_size()` and `size()`.
- Minor code cleanups and formatting for improved readability.

### 📦 Why this matters:
The test `wrap_word_fallback` was previously failing due to borrowing from a temporary value. This patch resolves that safely and makes the test easier to debug and maintain.

### DIFF
--- a/tests/wrap_word_fallback.rs
+++ b/tests/wrap_word_fallback.rs
@@ -1,38 +1,56 @@
-use cosmic_text::{Attrs, Buffer, FontSystem, Metrics, Shaping, Wrap};
+use cosmic_text::{Attrs, Buffer, Family, FontSystem, Metrics, Shaping, Wrap};
 
-// Tests the ability to fallback to glyph wrapping when a word can't fit on a line by itself.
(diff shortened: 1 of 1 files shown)


KushalMeghani1644:
just fixed the duplicate line problem as you mentioned. As of the unwrap() - I used .expect("Buffer width is not set") instead of .unwrap() to make debugging easier in case the test fails due to unexpected layout conditions. While we don’t expect this to be None, I find that a clear panic message helps catch regressions or misconfiguration faster — especially in test code where things evolve often. Happy to switch to .unwrap() if you'd prefer consistency, though.

KushalMeghani1644:
I have made some changes, and would love feedbacks on those! btw Thanks for giving a look at my PR, it's my first ever @grovesNL

PR #399, opened 2025-07-01
Do not ignore font family when querying fonts

### LINKED ISSUE #386: Why does font lookup completely ignore family?
Hey there, I was trying to use cosmic-text, setting up my desired fonts beforehand on `fontdb` and then passing the db to `FontSystem::new_with_locale_and_db`.

```rust
            const FONTS: [&'static [u8]; 2] = [
                include_bytes!("../res/whatever.ttf"),
                include_bytes!("../res/font2.ttf"),
            ];

            let mut db = fontdb::Database::new();
            for src in FONTS.map(Vec::from) {
                db.load_font_data(src);
            }

            // ... loading 5 different fonts, one for each family...
            db.set_serif_family("Whatever");
            db.set_fantasy_family("Font-2");

            let mut font_system = FontSystem::new_with_locale_and_db("en-US".to_string(), db);
```

Fundamentally the fonts I put into the db got picked up, but the selection via `Attrs::new().family(…)` on rendered text was completely haphazard. Meaning I would get random fonts depending on which fonts I installed, but seldom the font I had actually configured via `set...family("fontname")` on the db. Only monospace discernment seemed to _sometimes_ work.

I dug into the code a bit:
`fontdb` has a very nice and flexible `query` method with a very complex implementation that seems to do the right thing.
But cosmic-text doesn’t call that function at all, rather it selects a font from the DB based on a very simple custom logic that comes down to matching `Stretch` and `Style`, but not at all family.

https://github.com/pop-os/cosmic-text/blob/987ff45ff20f3cee1322e2f2909ac4c2c26f8321/src/font/system.rs#L326

https://github.com/pop-os/cosmic-text/blob/987ff45ff20f3cee1322e2f2909ac4c2c26f8321/src/attrs.rs#L323-L327

Is there a reason for this behaviour, and if yes, what is the recommended correct way to configure which fonts will get used?

Best regards

Joshix-1:
Fixes #386 

Font family now gets used when querying the font.
Not sure if this solution is good and if the old code is even needed. 
This works for me, which makes it good enough for me.

### DIFF
--- a/src/font/system.rs
+++ b/src/font/system.rs
@@ -6,6 +6,7 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::fmt;
 use core::ops::{Deref, DerefMut};
+use fontdb::Query;
 
 // re-export fontdb and rustybuzz
 pub use fontdb;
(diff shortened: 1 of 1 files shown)


ThePuzzlemaker:
This seems similar, but a bit more thought out, than what I did in [my fork](https://github.com/pop-os/cosmic-text/commit/f36730f121f0bc9af1047f4319e75e955c837aa1). The old code I believe is still necessary for further fallback behaviour beyond fontdb's basic font-family querying.

LGTM.

PR #400, opened 2025-07-02
Variable font support

ThePuzzlemaker:
Here's a pretty naïve solution for variable fonts.

The iterator doesn't use the match keys' weight, but instead tries to get the requested ideal weight, if the font is variable, otherwise it is ignored and the actual (non-variable) weight is used. This is because I didn't implement finding variable weight support for match keys; doing so would be impossible without parsing TTF files when matching and I didn't want to add that potentially expensive infrastructure if not entirely necessary.

This is a breaking change, and I'm open for ideas on how to fix that if it's an issue.

### DIFF
--- a/examples/rich-text/src/main.rs
+++ b/examples/rich-text/src/main.rs
@@ -24,6 +24,7 @@ fn set_buffer_text(buffer: &mut BorrowedWithFontSystem<'_, Buffer>) {
     let serif_attrs = attrs.clone().family(Family::Serif);
     let mono_attrs = attrs.clone().family(Family::Monospace);
     let comic_attrs = attrs.clone().family(Family::Name("Comic Neue"));
+    let inter_attrs = attrs.clone().family(Family::Name("Inter Variable"));
 
     let spans: &[(&str, Attrs)] = &[
         (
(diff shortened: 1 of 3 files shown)


ThePuzzlemaker:
N.B. I haven't tested whether or not this works with the monospaced font infrastructure in cosmic-text, but it seems passes all existing tests.

Existing monospaced fonts work alright, but I'm not sure about variable monospace fonts as I'm not sure if there are any really out there.

jackpot51:
The build CI job will have to pass before this can be merged

ThePuzzlemaker:
New commit should have fixed all clippy warnings/errors by code I touched and the merge conflicts.

Also removed the `.unwrap()` I added because it shouldn't fail, and if it does it's a recoverable error and not a panic.

jackpot51:
The build failure is due to formatting. If you want, I can push the changes for that, or you can run `cargo fmt`.

jackpot51:
This looks great now - it would be nice to see an example included using variable fonts. I will leave it up to you if you want to do that as part of this PR or a new one.

ThePuzzlemaker:
Added an example. Let me know if you need me to squash or change anything.

PR #402, opened 2025-07-06
Add pixel font flag

### LINKED ISSUE #397: Equivalent to Freetype's `LoadFlag::TARGET_MONO`?
I'm working on a small program to generate labels for my label maker - which only works with 1-bit images (a pixel is either on or off).

I wanted to use cosmic text for the layouting, but it doesn't seem capable of producing mono output. I tried the workaround that Bevy implemented here: https://github.com/bevyengine/bevy/pull/15368 but it still seems off compared to loading the font directly with FreeType and rendering the glyph with `LoadFlag::TARGET_MONO`.

Here's an image - the top is using Freetype, the bottom using cosmic text. It looks like a few of the letters have been shifted just enough to now spread over two pixels, or something similar.
The second image shows the text overlaid - cosmic text in red on top of the freetype.

Image

Image

callym:
This fixes #397 at least for the use-case in the issue (the example there now renders pixel-for-pixel the same as Freetype.

I'm not sure if this is a good fix, or if the maths is right, etc etc - any pointers or advice about how to implement this in a way that'll actually work/be useful/able to be merged would be super appreciated!

### DIFF
--- a/src/swash.rs
+++ b/src/swash.rs
@@ -34,7 +34,14 @@ fn swash_image(
 
     // Compute the fractional offset-- you'll likely want to quantize this
     // in a real renderer
-    let offset = Vector::new(cache_key.x_bin.as_float(), cache_key.y_bin.as_float());
+    let offset = if cache_key.flags.contains(CacheKeyFlags::PIXEL_FONT) {
+        Vector::new(
+            cache_key.x_bin.as_float().round() + 1.0,
+            cache_key.y_bin.as_float().round(),
+        )
+    } else {
+        Vector::new(cache_key.x_bin.as_float(), cache_key.y_bin.as_float())
+    };
 
     // Select our source order
     Render::new(&[
--- a/src/glyph_cache.rs
+++ b/src/glyph_cache.rs
@@ -9,6 +9,8 @@ bitflags::bitflags! {
         const FAKE_ITALIC = 1;
         /// Disable hinting
         const DISABLE_HINTING = 2;
+        /// Render as a pixel font
+        const PIXEL_FONT = 4;
     }
 }
 


callym:
The font in the linked issue is [NotJam Old Style 11](https://not-jam.itch.io/not-jam-old-style-11)

With another pixel font ([Undead Pixel 11](https://not-jam.itch.io/undead-pixel-light-8)):

`0.14.2` is one pixel offset, but doesn't display the weird shaping of the previous font:
image

Showing the whole alpha instead of using the Bevy alpha threshold hack shows the difference:
image

Pixel font flag set makes it pixel-perfect
image

With non-pixel fonts, this still doesn't match exactly but still seems closer than main's behaviour (although this isn't a pixel font so setting a pixel font flag doesn't make a lot of sense)

With `NotoSans-Regular`, this is `0.14.2`
image

image

With `NotoSans-Regular`, this is with this PR (and the pixel font flag set)
image

image

I think the actual solution to this issue needs to be somewhere else in the code/dependencies where the actual pixel hitting happens? Ideally there'd be an option to say you want to render with 1-bit alpha instead of 8-bit alpha, and then it'd do the calculations accordingly. But I'm not sure where that code is.

src/swash.rs
```diff
@@ -34,7 +34,14 @@ fn swash_image(
 
     // Compute the fractional offset-- you'll likely want to quantize this
     // in a real renderer
-    let offset = Vector::new(cache_key.x_bin.as_float(), cache_key.y_bin.as_float());
+    let offset = if cache_key.flags.contains(CacheKeyFlags::PIXEL_FONT) {
+        Vector::new(
+            cache_key.x_bin.as_float().round() + 1.0,
```
jackpot51, line 39: Why is 1.0 added here?

callym: I have no idea, but without it the font is shifted 1px to the left compared to the Freetype render I've been using as a comparison.

jackpot51: This is acceptable for now. I think in the future we will have a better way to deal with this - but it will require hinting changes at a lower level.

PR #403, opened 2025-07-08
Fixed Tab indenting the line instead of adding Tab or spaces

vancha:
Bug first reported here: pop-os/cosmic-edit#206

I think I've narrowed the issue down to cosmic-text rather than cosmic-edit. The current implementation assumed a selection, even where it could be None, and inserted tabs at the end of leading whitespace.

Re-using the old cursor position in cases where no selections had been made, inserts the tabs at the position where I think most users will expect them.

If this is a sub-par fix, I'll gladly change what is required.

### DIFF
--- a/src/edit/editor.rs
+++ b/src/edit/editor.rs
@@ -697,30 +697,34 @@ impl<'buffer> Edit<'buffer> for Editor<'buffer> {
                     self.with_buffer(|buffer| {
                         let line = &buffer.lines[line_i];
                         let text = line.text();
-                        // Default to end of line if no non-whitespace found
-                        after_whitespace = text.len();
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

vancha:
I hope this is a better solution, it calculates both the indentation width, and makes it hopefully clearer that there are two distinct paths. I would like to apologize for accidentally making a change to set_redraw by moving it out of the loop. That was unintentional, I'll put that back. Potential feedback appreciated if it's still not as expected.

jackpot51:
This looks good now, it just needs to pass the Rust/build check.

vancha:
Thank you, the problem seems to have been a styling error, which was solved by running rustfmt.

jackpot51 (approved):

Issue #332, opened 2024-12-09
[Feature request] Outlining text


sowelipililimute:
![grafik](https://github.com/user-attachments/assets/ba83fd4b-168c-49c4-b23a-70efb8e137c9)

It would be nice to be able to render an outline around text as pictured above.

kiedtl:
The PR #405 mentioned that it's better to do this downstream, but is this really out of scope for cosmic_text?

I'm also curious about other text effects, like shadows.

PR #405, opened 2025-07-19
basic text outlining

Joshix-1:
It's good enough for me for now, thought I would share it, if you find this interesting.

This should probably be done with more proper font rendering.

This algorithm isn't perfect:
    - it looks a bit off sometimes
    - it draws the same pixel multiple times

But with a thin outline on big text it looks fine

Issue: #332

### DIFF
--- a/src/attrs.rs
+++ b/src/attrs.rs
@@ -30,38 +30,44 @@ impl Color {
 
     /// Get a tuple over all of the attributes, in `(r, g, b, a)` order.
     #[inline]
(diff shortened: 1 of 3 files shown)


Joshix-1:
I think with the new Renderer trait, it's easier to implement this downstream.

https://codeberg.org/asozialesnetzwerk/john-of-us/src/commit/76ec0424eb5e4cd00e1dfb02c22827fa0a70db0e/crates/meme_creator/src/renderer.rs

PR #422, opened 2025-09-09
Experiment: Bump Peniko to use Linebender Resource Handle

DJMcNab:
This is not to be merged, but is instead being used to validate https://github.com/linebender/peniko/pull/129 by running your CI. The `ci.sh` script passes locally, but maybe that misses something that your actions CI catches.

Note that it's likely that Cosmic text will want to replace your Peniko usage with Linebender Resource Handle instead, once Peniko 0.4.1 is out

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,3 +87,6 @@ opt-level = 1
 
 [package.metadata.docs.rs]
 features = ["vi"]
+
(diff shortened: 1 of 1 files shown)


DJMcNab:
Thanks for approving the CI job. I don't think this needs any further action.

PR #433, opened 2025-11-01
Do not override default fonts

jackpot51:
@hecrj @nicoburns this may improve performance by letting fontdb select the default fonts per platform.


hecrj:
I get regressions even after uninstalling Open Sans.

Instead of trying to find Open Sans, `default_font_match_key` seems to always be querying for "FreeSans" now; which I don't have installed either :sweat_smile:

nicoburns:
This is identical to what I tested locally the other day, so I assume it will have the same performance improvement (equivalent to setting the font to DejaVuSans)

PR #450, opened 2026-01-01
Add automatic fake italic & fix fallback for fonts without true italic

TJYSunset:
When using a constrained font database, italic text without such fonts will appear as tofu:

Additionally, if no italic font exists in the database at all and a non-existent font family is specified, cosmic-text would panic at `cosmic-text\src\shape.rs:314:33: no default font found`. 

Most if not all CJK fonts don't have italics, but it's common for them to appear within an italicized formatting context such as a HTML `<em>` tag. This PR adds fake italic for such scenario. The previous text should look like this after the fix (together with #449):

Do note that this PR also makes almost every font eligible for fallback and might negatively impact performance in other situations. I also don't quite understand cosmic-text's internals, so this might break other things.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -546,14 +546,11 @@ fn override_fake_italic(
     font: &Font,
     attrs: &Attrs,
 ) -> CacheKeyFlags {
-    cache_key_flags
-        | if !font.italic_or_oblique
-            && (attrs.style == Style::Italic || attrs.style == Style::Oblique)
-        {
-            CacheKeyFlags::FAKE_ITALIC
-        } else {
-            CacheKeyFlags::empty()
-        }
+    if !font.italic_or_oblique && (attrs.style == Style::Italic || attrs.style == Style::Oblique) {
+        cache_key_flags | CacheKeyFlags::FAKE_ITALIC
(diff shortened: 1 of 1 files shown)


src/font/system.rs
```diff
@@ -17,11 +17,45 @@ use super::fallback::{Fallback, Fallbacks, MonospaceFallbackInfo, PlatformFallba
...
 }
 
+impl FontMatchKey {
+    fn new(attrs: &Attrs, face: &FaceInfo) -> FontMatchKey {
+        let not_emoji = !face.post_script_name.contains("Emoji");
+        let font_weight_diff = attrs.weight.0.abs_diff(face.weight.0);
+        let font_weight = face.weight.0;
+        let font_stretch_diff = attrs.stretch.to_number().abs_diff(face.stretch.to_number());
+        let font_stretch = face.stretch.to_number();
+        let font_style_diff = match (attrs.style, face.style) {
```
jackpot51: I don't see this value being used anywhere

TJYSunset: It's used by the derived `Ord` implementation, at L367 `font_match_keys.sort();`.

PR #452, opened 2026-01-09
fix: prevent line break opportunities from splitting ligatures   (e.g. |> != ++)

adam-r-kowalski:
## Issue
  Certain coding ligatures (specifically |> but also != and ++) fail to
  render correctly in cosmic-text because the unicode-linebreak crate treats
  characters like | (Vertical Line) as a break opportunity (class BA/AL).
  This causes ShapeSpan::build to split these sequences into separate
  ShapeWords. Since the shaping engine processes each word independently, it
  never sees the full sequence required to form the ligature.

## Fix
  This PR adds a check in ShapeSpan::build to manually inspect the
  characters at a proposed break point. If the sequence matches a known
  high-priority ligature (currently |>, !=, ++), the break opportunity is
  skipped, keeping the characters in the same shaping run.

## Verification
  Added a regression test test_ligature_segmentation in
  tests/shaping_and_rendering.rs that asserts these sequences remain as a
  single word after segmentation.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -8,6 +8,8 @@ use crate::{
     LayoutLine, Metrics, Wrap,
 };
 #[cfg(not(feature = "std"))]
+use alloc::format;
+#[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
 
 use alloc::collections::VecDeque;
@@ -854,6 +856,7 @@ impl ShapeSpan {
                             // 2. If we have the same number of glyphs, they might be contextual alternates (e.g. |> becoming 2 special glyphs).
                             // Check if the glyphs match the standard "cmap" (character to glyph) mapping.
                             // If they differ, the shaper substituted them, so we should keep them together.
+                            #[cfg(feature = "swash")]
                             if glyphs.len() == probe_text.chars().count() {
                                 let charmap = font.as_swash().charmap();
                                 let mut is_modified = false;


adam-r-kowalski:
Thanks for the feedback and for pointing out #392. I agree that hardcoding specific sequences wasn't
the right path.

I have pushed a new implementation that solves this dynamically without any hardcoded lists.

How it works:
Inside ShapeSpan::build, when we encounter a line break opportunity between two punctuation/symbol
characters, we perform a lightweight "probe":
1. We resolve the font for that specific pair.
2. We shape the pair (e.g., |> or !=).
3. We compare the resulting glyphs against the standard character-to-glyph mapping (cmap).
4. If the result is a single glyph (standard ligature) OR if the glyph IDs differ from the standard
mapping (contextual alternate), we detect a visual connection and skip the break opportunity.

Verification:
I've updated the test case to rely on the bundled Inter font.
* ->: Becomes 1 glyph (standard ligature). Kept together.
* |>: Becomes 2 glyphs, but with modified IDs (contextual alternate). Kept together.
* !=: Becomes 2 glyphs, but with modified IDs. Kept together.

This approach handles both standard ligatures (common in Fira Code) and multi-glyph contextual
alternates (common in JetBrains Mono or Inter) without manual lists.

adam-r-kowalski:
Screenshot 2026-01-16 at 9 39 46 PM

jackpot51:
@adam-r-kowalski thanks, I will be able to review soon.

jackpot51:
It looks like building with no_std is failing

adam-r-kowalski:
Thanks for catching that! I've pushed a fix to address the no_std build failures. I added the missing alloc::format import and guarded the swash-dependent ligature detection code so it only runs when the feature is enabled. I've verified that the no_std build passes locally now.

adam-r-kowalski:
It looks like the CI is failing on cargo deny because bincode has been flagged as unmaintained
  (RUSTSEC-2025-0141). This affects main as well (see recent run:
  https://github.com/pop-os/cosmic-text/actions/runs/21177607229/job/60910657584).

  This is coming from the syntect dependency:
  cosmic-text -> syntect -> bincode

  Since Cargo.lock is ignored, every CI run picks up this new advisory. bincode 1.3.3 is considered
  feature-complete/stable, so the likely fix is to explicitly ignore this advisory in deny.toml.

  Should I include that fix in this PR to get the checks passing?

jackpot51:
No, fixing that is not necessary.

PR #467, opened 2026-02-14
Ellipsize

hojjatabdollahi:
It's been almost exactly 3 years since I opened #84. wow!

It's time to finish the job.

This is a work in progress. Opened this PR to get feedback, and let @jackpot51 know it's pretty close to be done.

Here's the Enum:

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub enum Ellipsize {
    /// No Ellipsizing
    None,
    /// Ellipsizes the start of the sentence.
    /// IMPORTANT: Only works if `Wrap` is set to `Wrap::None`.
    Start,
    /// Ellipsizes the middle of the sentence.
    /// IMPORTANT: Only works if `Wrap` is set to `Wrap::None`.
    Middle,
    /// Ellipsizes the end of the sentence.
    /// If `Wrap` is `Wrap::None` then the `EllipsizeHeightLimit` is ignored and the ellipsis will
    /// be added at the end of the line.
    /// Otherwise the ellipsis will be added at the end of the last line that fits within the
    /// `EllipsizeHeightLimit`.
    End(EllipsizeHeightLimit),
}

#[derive(Debug, Clone, Copy, PartialEq)]
pub enum EllipsizeHeightLimit {
    /// Number of lines to show before ellipsizing the rest. Only works if `Wrap` is NOT set to
    /// `Wrap::None`.
    Lines(usize),
    //TODO: Height(f32) in pixels
}
```
Done:
- [x] `Ellipsize::Start`
- [x] `Ellipsize::Middle`
- [x] `Ellipsize::End`
- [x] `EllipsizeHeightLimit(Lines(usize))`
- [x] `EllipsizeHeightLimit(Height(f32))`
- [x] RTL, LTR, Mixed
- [x] Little visual bugs in bidi text (currently fixing this)
- [x] Tests

image
:partying_face:

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -1086,6 +1086,9 @@ struct VisualLine {
     spaces: u32,
     w: f32,
     ellipsized: bool,
+    /// Byte range (start, end) of the original line text that was replaced by the ellipsis.
+    /// Only set when `ellipsized` is true.
+    elided_byte_range: Option<(usize, usize)>,
 }
 
 impl VisualLine {
@@ -1094,6 +1097,7 @@ impl VisualLine {
         self.spaces = 0;
         self.w = 0.;
         self.ellipsized = false;
+        self.elided_byte_range = None;
     }
 }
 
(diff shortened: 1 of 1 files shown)


mmstick:
Perhaps ellipsizing in the middle should be the default since the start and end portion of a file name usually has the most important context.

hojjatabdollahi:
It looks like Gnome Files sets a height limit (let's say 2 lines) and then chooses "Ellipsize::Middle". Pango then ellipsizes the last visible line. My design only allowed for the last line to have End ellipsizing. 

So I spent today changing the API to allow for setting height limit and applying Start, Middle, and End ellipsizing to the last visible line.

image



The current API looks like this now:

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub enum Ellipsize {
    /// No Ellipsizing
    None,
    /// Ellipsizes the start of the last sentence that fits within the `EllipsizeHeightLimit`
    Start(EllipsizeHeightLimit),
    /// Ellipsizes the middle of the last sentence that fits within the `EllipsizeHeightLimit`.
    Middle(EllipsizeHeightLimit),
    /// Ellipsizes the end of the last sentence that fits within the `EllipsizeHeightLimit`.
    End(EllipsizeHeightLimit),
}

#[derive(Debug, Clone, Copy, PartialEq)]
pub enum EllipsizeHeightLimit {
    /// Number of lines to show before ellipsizing the rest. Only works if `Wrap` is NOT set to
    /// `Wrap::None`. Otherwise, it will be ignored and the behavior will be the same as `Lines(1)`
    Lines(usize),
    /// Ellipsizes the last line that fits within the given height limit. If `Wrap` is set to
    /// `Wrap::None`, the behavior will be the same as `Lines(1)`
    Height(f32),
}

```

I chose to set the height limit separate from the `buffer.height_opt` because we layout each line separately. So `EllipsizeHeightLimit::Height` is just measuring the height of one line (wrapped). It would be cool if we set the height limit based on buffer's height, but that was too much for one day! And I need some feedback to make sure I'm on the right track.

jackpot51:
@hojjatabdollahi the approach looks good to me, what do you have left before this is ready for review and merge?

hojjatabdollahi:
@jackpot51 
Here's what I'm thinking, you tell me which one I should do which one I can postpone, or you'll do it instead:

1- cleaning up the code (I have some changes already that I'll push in a couple of minutes)
2- performance regression tests (I did some and the percentages that I saw seemed too high, did some profiling, but couldn't figure out the culprit.
3- add tests
4- add an example (I have a GUI example that creates all permutations of Wrap, Align, and Ellipsize, but it's pretty slow, not sure if it's worth adding)

jackpot51:
Feel free to finish any in-progress changes. Other items could be done in a follow-up

hojjatabdollahi:
@jackpot51 

It's ready for review.

- I did not work on hit detection. We probably should add a property to LayoutRun and mark the line "truncated" or "ellipsized" because selection and hit detection should be handled differently for this otherwise clicking near the ellipsis causes a panic.
- Currently we consider the ellipsis LTR all the time which correctly shows the ellipsis in the middle of the line for Ellipsize::Middle. But Pango processes it like normal punctuation, which inherits the span direction, making it more readable. You can see the difference in these specific files (cosmic-files vs dolphin).

image

image

jackpot51:
Thanks, I will review tomorrow and try to fix the panic

hojjatabdollahi:
- I fixed the issue with different levels when ellipsizing RTL text in a LTR line.
image

Now it is acting similar to Pango.

I also committed some of my tests. I'll do more in a follow up.

Also I refactored and cleaned up the way I'm inserting ellipses. Previously the VisualLine contained ranges, and an EllipsizeState that would tell you in which index we should insert an ellipsis. And we would do that when creating the LayoutLine. But now, I added an EllipsisSpan and a magic index for it. And the ellipsis gets added to the visual line as a VlRange. When we're creating the LayoutLine if we encounter the magic index we return the ellipsis span. This helped simplify the last step and allow for reordering the ellipsis and making the final ellipsized text more readable and similar to Pango (the above image).

- I also think I fixed the panic that I mentioned earlier in 74b5260. But the code is not really tested and you may have a better idea how to fix that. 

Unless I encounter a huge bug, I won't make any more changes until you get a chance to review. Thank you.

jackpot51 (approved):

jackpot51:
@hojjatabdollahi now that this is merged, it is crucial to update iced and libcosmic at least with the minimal changes to compile. So far I see `iced/graphics/src/geometry/text.rs` needs ellipsize added on line 51.

PR #473, opened 2026-02-20
Remove completely unnecessary boxing/dyn of ranges introduced in latest changes.

dtzxporter:


hojjatabdollahi:
@dtzxporter thanks. I had a Rev<Range> that turned out to be unnecessary. Forgot to remove the boxes.

PR #476, opened 2026-02-20
shape: preserve cross-word contextual substitutions for ASCII LTR spans

johncalvinroberts:
Hello! First of all: thank you for this cool text layout library that works well in the `wgpu` ecosystem. 

Second: some context. I'm using `cosmic-text` alongside `glyphon` for some text rendering workloads with a bank of fonts, seeking to achieve rendering parity with a rendering system we're using on iOS/swift. I found some differences in the rendering using `cosmic-text` vs. our apple-native font rendering stack using `CATextLayer`. The discrepancy appears to be specifically with cross-span contextual alternatives. 

For example, this is with google fonts' `Mynerve`, there's a discrepancy.

On `CATextLayer`:
font_mynerve_regular ttf_expected

On `cosmic-text`:

font_mynerve_regular ttf_actual

If I'm understanding correctly, `cosmic-text` applies contextual alternatives (`calt`) based on word chunks, but we lose contextual alternatives if the span is multi-word or with line breaks due to this word chunking, leading to a discrepancy between native `CATextLayer` rendering and `glyphon` + `cosmic-text` rendering.

So, I came up with this diff with a bit of pair'ing with claude:
* Detect when `Shaping::Advanced` is used on a purely ASCII, LTR, multi-word span, shape the entire span first and then distribute the resulting glyphs into per-word buckets.
* Pretty conservative with the approach: add a `should_shape_with_span_context` guard to keep the fast/original per-word shaping path path for non-ASCII text, RTL runs, Shaping::Basic, and single-word spans.
*  Add a test verifying that glyph byte offsets land in the correct word buckets after the split, and that Shaping::Basic and single-word spans are unaffected.

Thank you for your esteemed code review. Let me know if any changes or more info is needed or a more runnable repro case is needed.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -792,6 +792,34 @@ impl ShapeWord {
     }
 }
 
+fn should_shape_with_span_context(
+    span: &str,
+    level: unicode_bidi::Level,
+    shaping: Shaping,
+) -> bool {
+    if shaping != Shaping::Advanced || level.is_rtl() {
+        return false;
+    }
+
+    let mut has_whitespace = false;
+    let mut has_non_whitespace = false;
+    for ch in span.chars() {
+        // Keep this path conservative to avoid changing behavior for complex
+        // scripts / controls where contextual shaping and bidi interactions are
+        // more subtle.
+        if !ch.is_ascii() || ch.is_ascii_control() {
+            return false;
+        }
+        if ch.is_whitespace() {
+            has_whitespace = true;
+        } else {
+            has_non_whitespace = true;
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


jackpot51:
Why would this be limited to ASCII?

johncalvinroberts:
> Why would this be limited to ASCII?

I know it wouldn't be explicitly limited to ascii. More implemented like this because I don't know what other locales would be appropriate or not to implement like this, so rather than blasting wide I implemented a narrow fix. Does applying cross-word contextual alternates apply to all locales? I will investigate this myself as well but would be great if you had an opinion you could share @jackpot51

Issue #312, opened 2024-09-03
Underlined text


abrisco:
Hi, thank you for this crate!

I was wondering, does this crate support underlined text?

Thanks!

abrisco:
Thank you @hojjatabdollahi!!

PR #480, opened 2026-02-24
Text Decoration

hojjatabdollahi:
Implements underline (single, and double lines), strikethrough, and overline.

Each decoration can have its own color (as opposed to one text_decoration_color config).
image

Boxed the decoration config to avoid bloating glyphs and impacting performance (adding 8 instead of 40 bytes).
Decorated texts are the cold path anyways.

note: the offset for `overline` should be based on glyph ascent. It needs more refactoring and since `overline` is rarely used I left that for a later time.

Closes https://github.com/pop-os/cosmic-text/issues/312.

### DIFF
--- a/src/attrs.rs
+++ b/src/attrs.rs
@@ -276,6 +276,9 @@ pub struct GlyphDecorationData {
     pub underline_metrics: DecorationMetrics,
     /// Strikethrough offset and thickness from the font
     pub strikethrough_metrics: DecorationMetrics,
+    /// Font ascent in EM units (ascent / upem).
+    /// Used for overline positioning
+    pub ascent: f32,
 }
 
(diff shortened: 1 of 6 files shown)


hojjatabdollahi:
@jackpot51 Here's the new approach:

- The user inputs a list of spans with different decorations
- Based on their direction and font, we merge them into spans
- For each span, we create a slice of decoration_spans (for the user-level spans that made up this span)
- Later when creating the LayoutLine, we look up the current glyph and find the decoration span that matches it

This should have no effect on text without decoration and negligible effect on text with decoration.

I tried to early exit where I can, or try to use a cursor when trying to map a glyph to it's decoration.

jackpot51 (approved):

Issue #280, opened 2024-07-04
Make reshaping/relayouting when inputs change lazy


nicoburns:
Currently when you call methods like `set_text`, `set_size`, and `set_metrics_and_size`, cosmic-text will eagerly reshape and relayout the Buffer. This is inefficient if multiple changes occur before the new text layout is required. For instance if both the text and the buffer size change, of if the buffer size changes multiple times within a frame.

I think we ought to change to setting a "dirty" flag (or flags - perhaps we could differentiate between when relayout and reshaping is required) when input parameters change. And then have a methods to relayout/reshape on demand which skip doing work if the flag isn't set.

Issue #308, opened 2024-09-02
Performance improvements


UkoeHB:
Areas for perf improvement:

- [ ] `Buffer::set_text` can be optimized in a similar fashion to #303.
- [ ] Use `SmallVec` instead of `Vec` where it makes sense. This should somewhat reduce the cost to make new buffers.
  - #311
- [ ] There are two allocations at the top of `set_rich_text` that can be moved to `ShapeBuffer`.
- [ ] Internally, `ShapeLine` is essentially a `Vec<Vec<Vec<ShapeGlyph>>>`. This can be flattened into two vecs: a vec of indices + metadata and a vec of ShapeGlyphs. Or even one vec with the metadata in-line.
  - I attempted this but got stuck on the need to sometimes reverse words and glyphs in `ShapeSpan`. It may be less efficient to reverse elements of a flattened vec than the current implementation.
- [ ] Internally, `BufferLine` caches a `Vec<LayoutLine>`, which is essentially `Vec<Vec<LayoutGlyph>>`. This could also be flattened.
  - The challenge here is needing/using `[LayoutGlyph]` in the layout iterator. We would no longer have access to a raw slice of glyphs.
- [x] `Buffer` caches a `ShapeBuffer` but it would be better to have this in `FontSystem` or elsewhere for global reuse.
  - #307
- [x] `FontFallbackIter::monospace_fallbacks` can be injected or reused from `FontSystem`
  - #309
- [ ] `shape_fallback()` returns a `Vec` that isn’t reused, and takes a `Vec` of glyphs. These can be moved to `ShapeBuffer`.
- [ ] `ShapeLine::adjust_levels` produces a `Vec` that can be cached.
- [ ] `ShapeLine::build` makes a new `unicode_bidi::BidiInfo` every time, which contains multiple allocations.
  - Sadly, `unicode-bidi` is not designed to let you reuse allocations. `wizterm-bidi` looks like it *might* be more performant, but the API is more restricted and I got stuck trying to test it out (it also doesn't support `no-std`).
- [ ] `ShapeLine::reorder` allocates twice internally.
- [ ] `BidiParagraphs` makes a new `unicode_bidi::BidiInfo` when constructed, which allocates multiple times.
- [ ] `ShapeLineKey` uses allocations that could be cached. It's also pretty expensive to hash.

PR #482, opened 2026-02-25
feat: builder method to apply changes to buffer in a batch

hojjatabdollahi:
Currently cosmic-text reshapes and relayouts with every method call. For example, set_text, set_wrap, set_ellipsize would trigger 3 separate layout calls.

This PR proposes using a builder method to apply changes to the buffer. The new API looks like this:

```rust
buffer.configure()
    .metrics(Metrics::new(18.0, 24.0))
    .size(Some(800.0), Some(600.0))
    .wrap(Wrap::Word)
    .text("Hello, world!", &attrs, Shaping::Advanced, None)
    .apply(&mut font_system);
```

- `apply()` determines the minimum amount of work needed.
- The old `set_*` methods are deprecated to signal the library users to switch to the new API. They now delegate to `configure()` internally.
- `set_text` is now reuses allocation. Previously only `set_rich_text` was doing that.

Fixes https://github.com/pop-os/cosmic-text/issues/280.
And some of the performance improvements mentioned in https://github.com/pop-os/cosmic-text/issues/308. I have more performance improvements but they're based on this branch, so I'll submit them later.

In my benchmarks (and logically) this PR either produces the same performance or improves performance.

I understand that an API change like this may be undesirable, but the alternative is making the `set_*` methods lazy, but then that would be a breaking change. (e.g. `set_wrap` doesn't need `font_system` and does nothing until `relayout` somehow is called and that should be conveyed to the user, which sounds more disruptive than deprecating the old methods). Also lazy-on-read would require `&mut self` since they `layout` mutates internal caches. 

If a breaking change is ok, then I can update the `set_*` methods to be lazy, and keep this builder method as sugar, or be removed.

I prefer the lazy approach, but this PR is a pragmatic solution given the current constraints.

### DIFF
--- a/benches/layout.rs
+++ b/benches/layout.rs
@@ -10,7 +10,7 @@ fn load_font_system(c: &mut Criterion) {
 fn layout(c: &mut Criterion) {
     let mut fs = ct::FontSystem::new();
     let mut buffer = ct::Buffer::new(&mut fs, ct::Metrics::new(10.0, 10.0));
-    buffer.set_size(&mut fs, Some(80.0), None);
+    buffer.configure().size(Some(80.0), None).apply(&mut fs);
 
     for (wrap_name, wrap) in &[
         ("None", ct::Wrap::None),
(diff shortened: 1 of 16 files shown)


jackpot51:
My guess about #280, but I cannot say, is that lifetime constraints would mean they would want to do multiple changes in different scopes before a final reshaping/relayout. It would be helpful of the people who referenced that issue from bevy and egui provided any feedback on their needs.

hojjatabdollahi:
superseded by https://github.com/pop-os/cosmic-text/pull/483

PR #503, opened 2026-04-14
fix: correct monospace width rounding for non-monospace fallback glyphs

ghost:
(_Note: AI was used to help find this fix. However, it's only six lines of code, and I understand what it does._)

The monospace width snapping logic was dividing x_advance (in pixels) by match_mono_em_width (cell_width / font_size), which produces a value in an intermediate unit space where one cell equals font_size rather than 1. This caused glyphs from non-monospace fallback fonts (e.g. Arabic, CJK from proportional system fonts) to snap to incorrect widths that were not integer multiples of the cell width.

For example, with font_size=14px and cell_width=8.4px:
- Old: 7.7px / 0.6 = 12.83, rounds to 13, 13 * 0.6 = 7.8px (wrong)
- New: 7.7px / 8.4 = 0.917, rounds to 1, 1 * 8.4 = 8.4px (correct)

The fix rounds x_advance / match_mono_width (pixel / pixel = number of cells) to get the correct cell count, then multiplies back by match_mono_width to get the snapped pixel advance. Visible glyphs (x_advance > 0) are guaranteed at least 1 cell so that narrow scripts like Sylheti Nagri don't collapse to zero width.

- [YES] I have disclosed use of any AI generated code in my commit messages.
- [YES] I understand these changes in full and will be able to respond to review comments.
- [YES] My change is accurately described in the commit message.
- [YES] My contribution is tested and working as described.
- [YES] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -2903,9 +2903,12 @@ impl ShapeLine {
                                     0.0
                                 },
                             );
-                            if let Some(match_em_width) = match_mono_em_width {
-                                // Round to nearest monospace width
-                                x_advance = ((x_advance / match_em_width).round()) * match_em_width;
+                            if let Some(mono_width) = match_mono_width {
+                                // Round to nearest multiple of the monospace cell width
+                                let cells = (x_advance / mono_width).round();
+                                // Ensure visible glyphs occupy at least one cell
+                                let cells = if x_advance > 0.0 { cells.max(1.0) } else { cells };
+                                x_advance = cells * mono_width;
                             }
                             if hinting == Hinting::Enabled {
                                 x_advance = x_advance.round();


jackpot51:
Please rewrite the commit message to be simpler

averyspencer:
> Please rewrite the commit message to be simpler

Done :)

hojjatabdollahi:
I really like the new Linux Kernel rule to put "Assisted-by" in the commit, instead of "Co-Authored-By" to clearly designate AI as a tool rather than a legal collaborator, ensuring human liability for the submitted code.

https://docs.kernel.org/process/coding-assistants.html#attribution

jackpot51:
With how simple the change has become, can it just be written without any LLM? COSMIC may end up having a policy blocking all LLM use for code, and even if not, it is what I would prefer.

averyspencer:
> I really like the new Linux Kernel rule to put "Assisted-by" in the commit, instead of "Co-Authored-By" to clearly designate AI as a tool rather than a legal collaborator, ensuring human liability for the submitted code.
> 
> https://docs.kernel.org/process/coding-assistants.html#attribution

I will fix this when I get home tonight.

averyspencer:
> With how simple the change has become, can it just be written without any LLM? COSMIC may end up having a policy blocking all LLM use for code, and even if not, it is what I would prefer.

I don't really think it matters. If you don't want AI assisted code I can just close the PR  and file an issue report if you guys want.

I found this bug during my own usage for a personal project, not from an issue report. I tested it and it fixes the problem. The patched version is installed on my system.

hojjatabdollahi:
I've been looking into this for a couple of days. And I've got a question.

> This caused glyphs from non-monospace fallback fonts (e.g. Arabic, CJK from proportional system fonts) to snap to incorrect widths that were not integer multiples of the cell width.

Here's an example of Farsi (same for Arabic) falling back to a proportional font:

master (clearly not following monospace width):
monospace_width_snap

this PR:
monospace_width_snap

Since each glyph has its own x offset, this causes the glyphs to overlap. And since it's rounding, glyphs that need 1.4 cells are being places in a single cell. If we `ceil` instead of `round`:

monospace_width_snap

The glyphs don't collide, and they're are integer multiples of monospace width.


That being said, if `monospace_fallback` feature of `cosmic-text` is enabled it will find fallback for those glyphs on most systems.

My question is, what was your use-case? Where does this normally happen? Would `ceil` work in that case?

averyspencer:
I was making an app with a TUI that needed to support a bunch of different scripts and I noticed that some of the scripts weren't monospace in Cosmic Terminal, causing ugly visual glitches. I didn't have this problem in GNOME Terminal. Sorry if this is the wrong place for this.

Also good catch on the ceil vs. round thing.

Also sorry about my cheekiness about AI, I shouldn't have been using SaaS slop to begin with. I'll try rewriting it without AI.

bkRodrigo:
I tested a local `cosmic-term` build with a similar `cosmic-text` patch, and it fixed the rendering issue I reported in https://github.com/pop-os/cosmic-term/issues/785.
The issue showed up in opencode’s bottom-left loader. After rebuilding `cosmic-term` with the fallback glyph advance snapped to the actual monospace cell width, the loader stopped shifting/misrendering.
The local change I tested was this:
```rust
if let Some(mono_width) = match_mono_width {
    // Snap fallback glyphs to whole terminal cells.
    let cells = (x_advance / mono_width).ceil();
    let cells = if x_advance > 0.0 { cells.max(1.0) } else { cells };
    x_advance = cells * mono_width;
}
```
I used `ceil()` rather than `round()` based on the earlier review feedback about avoiding under-allocation/overlap for glyphs wider than one cell.
So from my side, this PR appears to be targeting the right root cause for the opencode/COSMIC Terminal issue.

hojjatabdollahi:
@bkRodrigo would you mind testing with https://github.com/pop-os/cosmic-text/pull/507

ghost:
Hi folks, unfortunately I can't be the person vouching for this pull request anymore. I have decided to stop using "AI", because you shouldn't use "AI", and I am unfortunately far too busy at the moment to do any manual programming. Hopefully someone will get a human-written variation of this fix through regardless.

PR #507, opened 2026-05-27
fix: use unicode-width to get the correct width of monospace glyphs

hojjatabdollahi:
Supersedes https://github.com/pop-os/cosmic-text/pull/503

Alacritty uses unicode-width to create the grid it produces in cosmic-term. Currently we pass the text to cosmic-text and it lays it out using x-advance which for some glyphs it creates inconsistencies. The previous attempt was to round/ceil to a whole cell number, but turns out it's not that simple, the unicode-width crate is a better reference. So let's use that.

In this PR, we use the same library to see how many cells a monospace glyph should take. 

This doesn't (shouldn't) affect normal buffers (they don't have match_mono_width set), and for monospace buffer should only add a couple of nanoseconds. Most terminal buffers are just ASCII characters so they're short-circuited, and for the rest it's a quick static lookup, shouldn't add more than a couple of nanoseconds (compared to ms for shaping).

___

- [ ] My contribution is tested and working as described.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ sys-locale = { version = "0.3.2", optional = true }
 unicode-linebreak = "0.1.5"
 unicode-script = "0.5.8"
 unicode-segmentation = "1.12.0"
+unicode-width = "0.2.2"
 
 [dependencies.swash]
 version = "0.2.6"
--- a/src/buffer_line.rs
+++ b/src/buffer_line.rs
@@ -259,7 +259,8 @@ impl BufferLine {
                 .layout_opt
                 .take_unused()
                 .unwrap_or_else(|| Vec::with_capacity(1));
-            let shape = self.shape(font_system, tab_width);
+            self.shape(font_system, tab_width);
+            let shape = self.shape_opt.get().expect("shape not found");
             shape.layout_to_buffer(
                 &mut font_system.shape_buffer,
                 font_size,
@@ -270,6 +271,7 @@ impl BufferLine {
                 &mut layout,
                 match_mono_width,
                 hinting,
+                &self.text,
             );
             self.layout_opt.set_used(layout);
         }
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -10,6 +10,7 @@ use crate::{
 };
 #[cfg(not(feature = "std"))]
 use alloc::{format, vec, vec::Vec};
+use unicode_width::UnicodeWidthChar;
 
 use alloc::collections::VecDeque;
 use core::cmp::{max, min};
@@ -1588,6 +1589,7 @@ impl ShapeLine {
             &mut lines,
             match_mono_width,
             hinting,
+            "",
         );
         lines
     }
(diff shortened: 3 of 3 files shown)


willstott101:
I can confirm that this fixes the glyph width issues in opencode reported here: https://github.com/pop-os/cosmic-term/issues/785

**I am using "DejaVu Sans Mono" - top is what's installed on 24.04 - bottom is my build of cosmic-term with this branch:**
image

```
#!/usr/bin/env python3
REPEAT = 40

GLYPHS = {
    "ascii dot (reference)": ".",
    "too short: BLACK VERY SMALL SQUARE": "\u2B1D",
    "too wide: BLACK PENTAGON": "\u2B1F",
    "BLACK SQUARE": "\u25A0",
    "BALLOT X": "\u2717",
}

for name, c in GLYPHS.items():
    print(f"  {c * REPEAT}  {name}")
```

LeonardoPaccianiMori:
Supporting evidence for this approach from OpenCode's TUI on COSMIC Terminal 1.8.0 (Pop!_OS 24.04, Wayland, Noto Sans Mono). The current advance-based layout produces two visible failures here:

**1. Fallback glyph (colorized animated Unicode)**

U+2B1D (BLACK VERY SMALL SQUARE) is missing from common terminal monospace fonts, so it is shaped from a fallback face (FreeMono/FreeSans). Rapid colorized `\r` redraw leaves horizontal streaks/fragments in the animated row. A standalone repro (8 cells alternating `■` U+25A0 and `⬝` U+2B1D at 40 ms) reproduces it; replacing only `⬝` with `·` (U+00B7, present in the main font) removes the artifact. This matches pop-os/cosmic-term#785.

**2. Wide glyph in the main font**

`▣` U+25A3 is present in Noto Sans Mono but has a **2.0× cell advance**, while unicode-width reports 1 cell (EAW=A). A row containing it shifts everything after the glyph one column to the right, overwriting the adjacent sidebar:

```bash
printf '%s\n' '1234567890123456789012345678901234567890'
printf 'X  Build · DeepSeek V4.1 Flash        |\n'
printf '▣  Build · DeepSeek V4.1 Flash        |\n'
printf '◈  Build · DeepSeek V4.1 Flash        |\n'
printf '■  Build · DeepSeek V4.1 Flash        |\n'
printf '%s\n' '--------------------------------------|'
```

The trailing `|` is one column further right on the U+25A3 and U+25C8 rows (U+25C8 is also 2.0 cells in Noto Sans Mono).

Measurements (fontTools, hmtx advance / cell width 0.6 em):

- Noto Sans Mono: U+25A3 = 2.000×, U+25C8 = 2.000×; U+25A0, U+00B7, U+2022, U+25CF = 1.000×
- DejaVu Sans Mono: all 1× except U+2B1D missing

Both cases are "glyph occupancy disagrees with the terminal cell grid". Deriving the cell count from unicode-width (this PR) should cover both the fallback-glyph and the main-font-wide-glyph cases. Reports on the app side: anomalyco/opencode#48378 and anomalyco/opencode#48380, closed in favor of fixing this at the renderer level.

PR #510, opened 2026-06-10
feat: persistent font cache

hojjatabdollahi:
Similar to `fc-cache` this saves the results of the previous scan in a file in the default cache directory. 
It's gated by `std` and is opt-in. Shouldn't affect anything if you choose not to use it.
Tried to make it self-contained with no extra dependencies.

This significantly improves speed in libcosmic apps on systems with many fonts. It almost turns it into a constant number, and since it's shared between all cosmic-text consumers (all libcosmic apps) it's a pretty nice speed up for COSMIC.

Here's the profile of opening cosmic-term with around 3k fonts (goes from almost 2 seconds to 0.5 second):
Before: https://share.firefox.dev/3Q6pbEh
After: https://share.firefox.dev/4vELNuD

___

### DIFF
--- a/src/font/cache.rs
+++ b/src/font/cache.rs
@@ -1,0 +1,424 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Persistent on-disk cache for the parsed system-font index.
+//!
+//! [`fontdb::Database::load_system_fonts`] walks every system font directory and
+//! `mmap`s + parses each font file to populate per-face metadata. The result lives only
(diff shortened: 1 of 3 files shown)


hojjatabdollahi:
@jackpot51 any feedback on this PR? This seems like a big win to me, is there a reason this was not attempted before, am I missing something?

jackpot51 (approved):
This looks good to me. I was on vacation when the PR was created but happy to merge now

PR #511, opened 2026-06-11
Decouple decoration from shaping

airstrike:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,10 @@ harness = false
 name = "text_shaping_benchmarks"
 harness = false
 
+[[bench]]
+name = "decoration_toggle"
+harness = false
+
 [workspace]
 members = ["examples/*"]
 
--- a/benches/decoration_toggle.rs
+++ b/benches/decoration_toggle.rs
@@ -1,0 +1,35 @@
+use cosmic_text as ct;
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+
+fn bench_decoration_toggle(c: &mut Criterion) {
+    let mut fs = ct::FontSystem::new();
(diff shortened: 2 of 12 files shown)


hojjatabdollahi:
This looks good to me.

Not sure looking up the font metrics on relayout instead of shaping would affect performance. Since the font is fixed, there is no point in recalculating it for every line that has a decoration when layout is invalidated (for example when resizing a window). But as I said, I'm not sure if it's a meaningful impact.

Please double check everything with explicit LTR/RTL base directions after you rebase on master. Just in case. Like so: `buffer.set_direction(Direction::RightToLeft);`

jackpot51:
Looks like there are comflicts, after those are resolved I can merge

airstrike:
Thank you both! I'll do a little bit more testing and rebase but it might be a couple of days before I can get to it.

airstrike:
Rebased on `main`. Thanks to @hecrj, who did the rebase and added a follow-up commit that moves the per-span decoration data into the reusable ShapeBuffer scratch instead of allocating a fresh `Vec` per layout. On the PR's `decoration_toggle` bench that's about 3% faster.

@hojjatabdollahi re explicit base directions: I added tests/decoration_direction.rs, which forces Direction::LeftToRight and Direction::RightToLeft via set_direction and checks that decoration spans still cover exactly the underlined glyphs with matching x extents. It covers Hebrew and English, single-line and wrapped, plus an underline crossing a Latin/Hebrew boundary, and each test asserts the forced direction actually took effect. Rendering of the same scenarios attached.

Please let me know if you need anything else.

image

airstrike:
cc @jackpot51 if you want to merge

PR #519, opened 2026-07-07
fix: honor cursor Affinity when resolving the caret position

abdu-benayad:
## The bug

`hit()` and the cursor motions set `Affinity::Before`/`After` carefully, but `LayoutRun::cursor_glyph` never reads it: at a boundary between runs of opposite direction the caret always resolves to the logical-after side. In an editor over a mixed bidi line the caret visually teleports across the direction seam regardless of which side the cursor arrived from — clicking at the end of the LTR half paints the caret in the middle of the RTL half (and vice versa).

## The fix

`Affinity::Before` now prefers the glyph that **ends** at the cursor index (returning its trailing edge), falling through to the existing start-of-glyph search when no glyph ends there (index 0). `Affinity::After` behavior is unchanged.

In single-direction text, end-of-glyph-N and start-of-glyph-N+1 are the same x, so this changes nothing outside mixed lines — the second test pins that.

## Tests

- `affinity_picks_the_caret_side_at_a_direction_seam` — "abc" + Arabic, cursor at the seam: fails before this change (`Before` resolved to the After-side glyph deep in the Arabic run), passes after.
- `affinity_is_moot_inside_a_single_direction_run` — both affinities agree in plain LTR text, before and after the change.

Found while building an RTL-first text stack on cosmic-text; part of a small series of RTL correctness fixes with #520 and #521.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -146,9 +146,27 @@ impl LayoutRun<'_> {
     /// Find which glyph in this run contains `cursor`, returning
     /// `(glyph_index, pixel_offset_within_glyph)`, or `None` if the cursor
     /// is not on this run.
+    ///
+    /// At an exact glyph boundary the cursor's [`Affinity`] picks the side:
+    /// `Before` attaches to the glyph that *ends* at the index, `After` to the
+    /// glyph that *starts* there. The two positions only differ where runs of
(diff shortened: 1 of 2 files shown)


jackpot51:
The PR template is mandatory

abdu-benayad:

PR #520, opened 2026-07-07
fix: hit() before-line detection for aligned lines

abdu-benayad:
## The bug

The clicked-before-the-line test for LTR runs in `Buffer::hit` is hardcoded `x < 0.0`, which is only correct for flush-left layout. Under `Align::Right` or `Align::Center` a short line starts at `glyphs[0].x > 0`, so a click in the leading gap (left of the text, but at x ≥ 0) falls through to the past-all-glyphs arm — placing the cursor at the line **end** instead of the start. In an editor, clicking just left of a right-aligned line jumps the caret to the wrong side of the line.

## The fix

Compare against `glyphs[0].x` — the true visual start of the run by L2-reorder construction — instead of 0.0. The RTL arm already does the equivalent (`x > glyph.x`).

## Tests

- `click_before_a_right_aligned_line_hits_its_start` — fails before this change (the click landed at the line end), passes after.
- `click_before_a_flush_left_line_still_hits_its_start` — pins the negative-x case the old test did handle.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1199,7 +1199,12 @@ impl Buffer {
                 'hit: for (glyph_i, glyph) in run.glyphs.iter().enumerate() {
                     if first_glyph {
                         first_glyph = false;
-                        if (run.rtl && x > glyph.x) || (!run.rtl && x < 0.0) {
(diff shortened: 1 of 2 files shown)


abdu-benayad:

PR #521, opened 2026-07-07
fix: fresh-line sentinel still aliases span 0 in width and overflow math

abdu-benayad:
## The bug

Follow-up to 4fe1195e, which guarded the tail-commit arm of `layout_spans` against the fresh-line `SpanWordGlyphPos::ZERO` sentinel aliasing a genuine resume at span 0 — that fixed the dropped glyphs, but the same aliasing survives at the other `== start.span` / `== start.word` comparison sites:

- **Width undercount:** word 0 of an incongruent fresh-line span measures **0 wide** through `get_glyph_start_end` (the aliased `start` makes it return the empty glyph range `(0, start.glyph = 0)`). The overflow check then believes more content fits than does, and an ellipsized line paints past the buffer width. Reproduced in the new test: forced-LTR base + RTL-first content ("سلام hello world"), `Ellipsize::End`, width at 60% of the full extent → the rendered extent exceeds the width by ~25%.
- **Overflow range truncation:** the overflow arm's `range_end` (`if span_index == start.span && !congruent { start.word_glyph_pos() }`) truncates to `(0, 0)` on a fresh line.

## The fix

Replace the ZERO sentinel with `span/word = usize::MAX`, which matches no real span or word — every resume arm is uniformly dead on fresh lines instead of each site needing its own guard. Traversal still starts at span 0 via an explicit `traverse_from` floor, and `remaining_content_exceeds` receives that floor rather than the sentinel (its Backward remaining-spans range needs the real lower bound, 0 on a fresh line). The now-unused `ZERO` const is removed.

The 4fe1195e guard and its `forced_ltr_keeps_rtl_glyphs` test keep working unchanged (the guard now only ever sees genuine resumes).

## Tests

- `ellipsis_respects_the_width_when_the_line_opens_incongruent` — fails before this change (extent 76.9 vs width 61.3 in this environment), passes after.
- Full existing suite green, including the direction and draw tests.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -1242,11 +1242,8 @@ struct SpanWordGlyphPos {
 }
 
 impl SpanWordGlyphPos {
-    const ZERO: Self = Self {
-        span: 0,
-        word: 0,
-        glyph: 0,
-    };
+    // NOTE: the old `ZERO` sentinel was removed; a fresh line (no resume point)
+    // now uses an unaliasable usize::MAX sentinel in `layout_spans` (see there).
     fn word_glyph_pos(&self) -> WordGlyphPos {
         WordGlyphPos {
             word: self.word,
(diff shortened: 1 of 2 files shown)


abdu-benayad:

PR #524, opened 2026-07-22
fix: color emoji fallback

### LINKED ISSUE #327: Inconsistent emoji rendering

jotuel:

Fixes #327

Emoji (☕, ☀️, ❤️ & ➡️) render inconsistently some colored, some monochrome. Symbol fonts earlier in the fallback chain ship monochrome outline glyphs for these codepoints and shape_fallback treats any non-zero glyph ID as covered so it never reached the color emoji font. Adds emoji-aware, upgrade-only, fallback selection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Prefer color font for emoji presentation clusters during fallback selection, fixing inconsistent (mixed color/monochrome) emoji rendering: https://github.com/pop-os/cosmic-text/issues/327
+
 ## [0.19.0] - 2026-04-22
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


jotuel:
I hope my attempt at slightly optimizing went better this time around. At least I learned that benchmarking on a laptop is challenging. Not sure if introducing smallvec dependency worth the miniscule gains. Can easily be reverted.

PR #532, opened 2026-08-26
swash: add opt-in LCD subpixel mask rasterization

raggi:
Add opt-in subpixel (LCD) glyph masks, selected per cache key with CacheKeyFlags::SUBPIXEL_RGB or CacheKeyFlags::SUBPIXEL_BGR. The flags name the physical stripe order they assume; the emitted mask stores channels in that order so consumers can composite without a swizzle.

The subpixel mask is produced by a three-stage path in swash_image:

1. Outline rasterization at 3x horizontal resolution, post hinting. This essentially matches FreeType's LCD mechanism.

2. Apply a 5-tap FIR filter matching the common FreeType "lcddefault". This filter reduces chromatic fringing.

3. Pack into RGBA: R/G/B hold filtered coverage for the physical stripe channels, A = (R + G + B) / 3 rounded. Coverage values stay linear and color-independent, easiest for arbitrary use in downstream composition.

Color glyphs keep the standard source order and still emit Content::Color, so the flags select a subpixel mask only where there is an outline to rasterize.

with_pixels previously dropped subpixel glyphs with a warning, since its Color-per-pixel API cannot carry per-channel coverage. Emit those glyphs using the mask's baked-in mean-coverage alpha instead, mirroring the Content::Mask arm: software renderers now render subpixel-keyed glyphs as grayscale AA rather than not at all. The existing arms are untouched, including their open base-alpha TODO, which the new one shares.

Zeno's subpixel implementation was avoided. Zeno's Format::Subpixel renders BGR when RGB is wanted and vice versa, never writes the alpha byte, applies no anti-fringing filter, and measures substantially slower than this path at 32px.

Updates #140: the patch here supplies the plumbing for strong subpixel rendering as we're providing a mask that is a similar quality to the stronger outputs of FreeType. Later patches to cyroglyph and iced must all be composed correctly to match / be superior to state of the art peers, as proper composition is critical.

Fixes #323 as much as is possible locally, there's a note in #323 to address configurability, and patches to do so will follow in libcosmic and cosmic-settings after the aforementioned glyphon and iced patches that will enable a system wide configuration of some of these parameters with sensible defaults.

AI was used in test scaffolding, several optimization and review passes. The code has been repeatedly human reviewed.


charliechensz:
## Field report: library-level works, end-to-end rendering is blurry (glyphon doesn't consume SubpixelMask yet)

I applied this patch to cosmic-text and wired it into cosmic-term, and wanted to share what I found — including a concrete blocker on the consumer side that this PR will need to address to actually improve rendering.

### What I did

1. **Patched cosmic-text** with this PR's changes (`CacheKeyFlags::SUBPIXEL_RGB/BGR` + `lcd_filter_downsample`), ported onto the crates.io 0.19.0 release (note: crates.io 0.19.0's `BufferLine::layout()` still takes 8 args — no `Direction` — while current main takes 9; the `swash.rs`/`glyph_cache.rs` hunks apply cleanly either way).
2. **Wired the flag in cosmic-term** (`src/terminal.rs`, both `Attrs::new()` call sites get `.cache_key_flags(CacheKeyFlags::SUBPIXEL_BGR)`), matched to the monitor's physical BGR stripe order (verified from EDID).
3. **Built cosmic-term 1.8.0** with the patched cosmic-text (release, 2m03s).

### Library-level verification: the patch works correctly

A direct `SwashCache::get_image_uncached()` call with `SUBPIXEL_BGR` on the CJK glyph 中 at 48px produces exactly what the PR describes:

- `content == SwashContent::SubpixelMask` (4 bytes/pixel RGBA vs 1 byte/pixel for the `Mask` baseline)
- channel separation on stroke edges matches the requested stripe order (B-strong fringes for BGR), 16.2% of inked pixels show >30 channel delta vs 0.0% for grayscale
- placement geometry is handled correctly by `lcd_filter_downsample`

So the rasterization side is solid.

### End-to-end result: blurry — root cause is on the consumer side

Rendering the same text in cosmic-term, the result is **visibly blurry** (worse than the existing grayscale AA). The cause is not in this PR — it's that cosmic-term's text renderer (glyphon) does not consume `SubpixelMask`:

```rust
// pop-os/glyphon, src/text_render.rs
SwashContent::SubpixelMask => {
    // Not implemented yet, but don't panic if this happens.
    ContentType::Mask
}
```

The subpixel mask is RGBA (each channel = coverage of one physical stripe); treating it as a 1-channel `Mask` samples/atlas it wrong, which produces the blurry output I observed.

### Suggestion

For this PR to land and actually fix rendering (the point of #363), the **consumer side needs a matching change**: glyphon's `text_render.rs` must map `SubpixelMask` to a proper RGBA content type (`ContentType::Color`-like path with subpixel blending), not fall through to `Mask`. That, plus any LCD filtering the compositor wants, is where the remaining work is.

Happy to help test a combined cosmic-text + glyphon change if someone wants to pick that up — the port/build setup (local git deps + `[patch]`) is reproducible in a couple of minutes.

